Issue #352, opened 2026-04-14
feat(llm): add openai-compat provider for locally-hosted LLMs (LM Studio, Ollama, llama.cpp)


letterj:
## Summary

Add an `openai-compat` LLM provider to the runtime that enables agents to run against locally-hosted models via any OpenAI-compatible `/v1/chat/completions` endpoint — LM Studio, Ollama (OpenAI-compat mode), llama.cpp server, vLLM — without requiring xAI credentials and without patching `xai-strict-filter.ts` to allowlist non-xAI model IDs.

## Problem

Currently the only way to run an agent against a local LLM on Apple Silicon is to add the model ID to `DOCUMENTED_XAI_MODEL_IDS` in `xai-strict-filter.ts` and route it through the Grok adapter. This:

- Corrupts the xAI strict filter, which exists specifically to enforce the documented xAI catalog
- Ties a local-model config to `apiKey` semantics that do not apply
- Breaks silently if the model ID is later removed from the allowlist patch
- Offers no startup validation that the local server is actually reachable

The `ollama` provider uses the `ollama` SDK, which works against Ollama's native API but not against LM Studio or llama.cpp, which expose only the OpenAI-compatible `/v1/chat/completions` interface.

## Current State

- `runtime/src/llm/grok/adapter.ts` — GrokProvider, xAI-specific, not appropriate for local models
- `runtime/src/llm/ollama/adapter.ts` — OllamaProvider, uses `ollama` SDK, does not work against LM Studio or OpenAI-compat endpoints
- `runtime/src/gateway/llm-provider-manager.ts` — `createSingleLLMProvider()` switch currently handles only `"grok"` and `"ollama"`; unknown providers return `null` with a warning
- `runtime/src/gateway/types.ts` — `GatewayLLMConfig["provider"]` union must be extended to include the new provider string

## Proposed Config Shape

```json
{
  "llm": {
    "provider": "openai-compat",
    "model": "google_gemma-4-26b-a4b-it",
    "baseUrl": "http://127.0.0.1:1234/v1",
    "apiKey": "local",
    "contextWindowTokens": 32768,
    "maxTokens": 4096
  }
}
```

- `baseUrl` — required; validated on startup to resolve to a local or LAN address (blocks accidental routing to external services)
- `apiKey` — required by the OpenAI SDK but not validated by local servers; any non-empty string works
- `contextWindowTokens` — required for prompt budget; local servers do not expose context window via a standard API field
[...]

PR #402, opened 2026-04-16
feat(llm): add openai-compat provider for locally-hosted LLMs

letterj:
_feat(llm): add openai-compat provider for locally-hosted LLMs_Closes #352

## Summary

Adds an `openai-compat` LLM provider that targets any OpenAI-format inference server (LM Studio, llama.cpp, vLLM, etc.) without requiring xAI credentials. Enables local/LAN model inference with the full AgenC agent runtime.

## Changes

**New files:**
- `runtime/src/llm/openai-compat/types.ts` — `OpenAICompatProviderConfig` interface + two typed error classes (`OpenAICompatConfigError`, `OpenAICompatConnectionError`)
- - `runtime/src/llm/openai-compat/openai-compat-filter.ts` — three pure validators: URL locality check, server reachability (`GET /v1/models`), model presence
- - `runtime/src/llm/openai-compat/adapter.ts` — `OpenAICompatProvider` class implementing the `LLMProvider` interface
- - `runtime/src/llm/openai-compat/adapter.test.ts` — 40 unit tests (openai SDK mocked via `vi.mock`)
- - `runtime/src/llm/openai-compat/openai-compat-filter.test.ts` — 41 unit tests
- - `runtime/src/llm/openai-compat/adapter.live.test.ts` — 2 live integration tests (skipped in CI, require `OPENAI_COMPAT_BASE_URL` + `OPENAI_COMPAT_MODEL`)
**Modified files:**
- `runtime/src/gateway/types.ts` — added `"openai-compat"` to the provider union type
- - `runtime/src/gateway/llm-provider-manager.ts` — registered provider instantiation; `contextWindowTokens` defaults to `32768` (not 4096) because the AgenC system prompt exceeds 14K tokens
- - `docs/RUNTIME_API.md` — new module entry, minimal config example, Local Inference profile, 32768 minimum note
## Design decisions

- **`contextWindowTokens` required in config** — local servers don't expose context size via API. Default fallback is 32768.
- - **Tool call IDs use `crypto.randomUUID()`** — fixes the Ollama-style duplicate-ID bug where the tool name was reused as the call ID across turns.
- - **Three startup checks** — baseUrl must be local/LAN, server must respond to `GET /v1/models`, requested model must be present in the response.
- - **Live tests in a separate file** — `vi.mock("openai")` hoisting in `adapter.test.ts` replaces the real SDK; co-locating live tests would break them.
## Minimal config

```json
{
  "llm": {
    "provider": "openai-compat",
    "model": "google_gemma-4-26b-a4b-it",
    "baseUrl": "http://127.0.0.1:1234/v1",
    "apiKey": "local",
    "contextWindowTokens": 32768
  }
}
```

## Test results

- 81 unit tests passing, 0 failures in new files
- - 2 live tests passing against LM Studio 0.4.9 / Gemma 4 26B
- - 0 regressions (19 pre-existing upstream failures, none in changed files)

### DIFF
--- a/docs/RUNTIME_API.md
+++ b/docs/RUNTIME_API.md
@@ -376,7 +376,7 @@ catalog: the server's `GET /v1/models` response is the authoritative source of t
 ```
 
 **Notes:**
-- `contextWindowTokens` is required — local servers do not expose context window size via API. Set this to match the model's actual context window (e.g. 32768 for Gemma 4 26B).
+- `contextWindowTokens` is required — local servers do not expose context window size via API. The AgenC system prompt alone requires more than 14K tokens, so 32768 is the recommended minimum. Values below 16384 will likely cause failures before the user sends a single message.
 - `apiKey` is passed in the `Authorization` header but is not validated by local servers. Any non-empty string works.
 - When running untrusted models, set `workspace.hostPath` to a neutral directory to prevent autonomous file access during initial context loading.
 - LM Studio: set the context window in the model settings to match `contextWindowTokens`. The LM Studio default of 4096 is too small for most system prompts.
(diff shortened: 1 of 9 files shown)


7etsuo:
Merged — thanks. The local/LAN startup validation and the UUID-based tool-call IDs are both solid.

PR #418, opened 2026-04-17
fix(llm): generate unique tool call IDs in Ollama adapter

### LINKED ISSUE #417: bug(llm): Ollama adapter reuses tool name as call ID causing duplicate ID errors on multi-step tool tasks
## Description

The Ollama LLM adapter assigns the tool function name as the tool call ID instead of generating a unique ID per invocation. When the same tool is called more than once in a multi-step sequence, the runtime receives duplicate tool call IDs and rejects the turn with:

```
Invalid tool-turn sequence (assistant_tool_call_id_duplicate) at message[N]:
assistant tool_calls contains duplicate id "<tool_name>"
```

Single-turn tool calls work fine. The bug manifests immediately on multi-step tasks (e.g. write → compile → run using `system.bash`).

## Affected code

Two paths in `runtime/src/llm/ollama/adapter.ts`:

- **Streaming path** (~line 362): `tc.function?.name ?? \`call_${toolCalls.length}\``
- - **Non-streaming `parseResponse`** (~line 678): `tc.function?.name ?? \`call_${i}\``
Both fall back to the tool name or a non-unique index-based string instead of a UUID.

## Reproduction

Configure the Ollama provider with any model that supports tool calling (e.g. `qwen2.5:14b` on an M2). Run any multi-step task that invokes the same tool twice in one session.

```
Invalid tool-turn sequence (assistant_tool_call_id_duplicate) at message[2]:
assistant tool_calls contains duplicate id "system.bash"
```

## Fix

Replace both ID assignments with `crypto.randomUUID()` — the same fix applied to the `openai-compat` provider in PR #402.

## Tested on

- Model: `qwen2.5:14b`, M2 32GB
- - LM Studio 0.4.9 / Gemma 4 26B (`google_gemma-4-26b-a4b-it`)

letterj:
Closes #417

## Summary

The Ollama adapter assigned the tool function name as the tool call ID instead of generating a unique ID per invocation. When the same tool was called more than once in a multi-step sequence, the runtime received duplicate tool call IDs and rejected the turn with `assistant_tool_call_id_duplicate`.

This is the same bug fixed in PR #402 for the `openai-compat` provider.

## Changes

**`runtime/src/llm/ollama/adapter.ts`** (+3 lines, -2 lines):
- Streaming path (~line 362): replaced `tc.function?.name ?? \`call_${toolCalls.length}\`` with `crypto.randomUUID()`
- - Non-streaming `parseResponse` (~line 678): replaced `tc.function?.name ?? \`call_${i}\`` with `crypto.randomUUID()`
- - Added `import crypto from "node:crypto"` at top
**`runtime/src/llm/ollama/adapter.test.ts`** (+60 lines):
- Two new regression tests: assert that two same-named tool calls get different IDs, and that neither ID equals the tool name
- - Both streaming and non-streaming paths covered
## Test results

- 30/30 Ollama adapter tests passing (28 existing + 2 new)
- - 21 pre-existing upstream failures on `upstream/main` — none in changed files, confirmed identical before and after this change
- - 0 regressions introduced
## Reproduction

Configure the Ollama provider with any tool-capable model (e.g. `qwen2.5:14b`). Run any multi-step task that invokes the same tool twice:

```
Invalid tool-turn sequence (assistant_tool_call_id_duplicate) at message[2]:
assistant tool_calls contains duplicate id "system.bash"
```

Single-turn tool calls were unaffected. Multi-step tasks (write → compile → run) failed immediately.

### DIFF
--- a/runtime/src/llm/ollama/adapter.test.ts
+++ b/runtime/src/llm/ollama/adapter.test.ts
@@ -157,6 +157,66 @@ describe("OllamaProvider", () => {
     expect(result.finishReason).toBe("tool_calls");
   });
 
+  it("assigns unique IDs to duplicate-named tool calls (non-streaming)", async () => {
+    const response = makeResponse({
+      message: {
+        content: "",
+        role: "assistant",
+        tool_calls: [
+          { function: { name: "search", arguments: { q: "first" } } },
+          { function: { name: "search", arguments: { q: "second" } } },
+        ],
+      },
+    });
+    mockChat.mockResolvedValueOnce(response);
+
+    const provider = new OllamaProvider({});
+    const result = await provider.chat([{ role: "user", content: "test" }]);
+
+    expect(result.toolCalls).toHaveLength(2);
+    expect(result.toolCalls[0].name).toBe("search");
+    expect(result.toolCalls[1].name).toBe("search");
+    expect(result.toolCalls[0].id).not.toBe(result.toolCalls[1].id);
+    expect(result.toolCalls[0].id).not.toBe("search");
+    expect(result.toolCalls[1].id).not.toBe("search");
+  });
+
+  it("assigns unique IDs to duplicate-named tool calls (streaming)", async () => {
+    const chunks = [
+      {
+        message: {
+          content: "",
+          role: "assistant",
+          tool_calls: [
+            { function: { name: "search", arguments: { q: "first" } } },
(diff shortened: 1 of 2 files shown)


7etsuo:
Merged — thanks for the Ollama fix and the duplicate-name regression tests.

PR #454, opened 2026-04-18
fix(gateway): expand tilde in keypairPath config field

### LINKED ISSUE #453: bug(gateway): tilde (~) not expanded in keypairPath config field
## Description

The `keypairPath` field in `~/.agenc/config.json` does not expand the tilde (`~`) to the user's home directory. When a config specifies:

```json
{
  "connection": {
    "keypairPath": "~/.config/solana/id.json"
  }
}
```

The runtime passes the literal string `~/.config/solana/id.json` to the filesystem, which fails with a file not found error because `~` is a shell shorthand, not an OS-level path construct.

## Affected file

`runtime/src/gateway/wallet-loader.ts`

## Reproduction

Set `keypairPath` to a tilde-prefixed path in config:

```json
{ "connection": { "keypairPath": "~/.config/solana/id.json" } }
```

Start the daemon — wallet fails to load with a file not found error. Using the absolute path (e.g. `/Users/username/.config/solana/id.json`) works correctly.

## Fix

Expand `~` to `os.homedir()` when resolving `keypairPath` before passing to the filesystem. The README and config docs show tilde paths as examples, so users will naturally use this format.

```typescript
import os from "os";
const expanded = keypairPath.startsWith("~/")
  ? path.join(os.homedir(), keypairPath.slice(2))
  : keypairPath;
```

letterj:
Closes #453

## Summary

The `keypairPath` field in `~/.agenc/config.json` did not expand the tilde (`~`) to the user's home directory. The runtime passed the literal string to the filesystem, causing a file not found error. The README and config docs show tilde paths as examples, so users will naturally use this format.

## Changes

**`runtime/src/gateway/wallet-loader.ts`** (+4 lines, -1 line):
- Added `import os from "node:os"` and `import path from "node:path"`
- - Added exported `expandPath(p)` helper: expands `~/` → `os.homedir()` + rest, passes everything else through unchanged
- - Wrapped `keypairPath` resolution with `expandPath()`
**`runtime/src/gateway/wallet-loader.test.ts`** (new file, +34 lines):
- 5 unit tests covering: `~/` expansion, nested `~/` path, absolute path passthrough, relative path passthrough, bare `~` passthrough (only `~/` is expanded)
## Test results

- 5/5 targeted tests passing
- - TypeScript: 0 errors

### DIFF
--- a/runtime/src/gateway/wallet-loader.test.ts
+++ b/runtime/src/gateway/wallet-loader.test.ts
@@ -1,0 +1,18 @@
+/**
+ * wallet-loader tests.
+ *
+ * expandPath has moved to runtime/src/types/wallet.ts — see wallet.test.ts.
+ */
+
+import { describe, it, expect } from "vitest";
+import * as walletLoader from "./wallet-loader.js";
+
+describe("wallet-loader module", () => {
+  it("exports loadWallet", () => {
+    expect(typeof walletLoader.loadWallet).toBe("function");
+  });
+
+  it("does not export expandPath (moved to wallet.ts)", () => {
+    expect((walletLoader as Record<string, unknown>).expandPath).toBeUndefined();
+  });
+});
--- a/runtime/src/types/wallet.test.ts
+++ b/runtime/src/types/wallet.test.ts
@@ -16,6 +16,7 @@ import {
   getDefaultKeypairPath,
   loadDefaultKeypair,
   KeypairFileError,
+  expandPath,
   Wallet,
 } from "./wallet";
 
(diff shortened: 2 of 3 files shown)


7etsuo:
Thanks, this is solid on the primary path. Before merge, could you move `expandPath` into `loadKeypairFromFile` in `runtime/src/types/wallet.ts` instead? Every keypair-loader call site (gateway WebSocket handlers + all the CLIs — `agent-cli`, `marketplace-cli`, `registry-cli`, `security`, `health`) funnels through that one function, so putting the expansion there closes #453 for every entry point in one change instead of one at a time. Happy to merge once that's in.

letterj:
Done — moved expandPath into runtime/src/types/wallet.ts and applied it at the entry of both loadKeypairFromFile and loadKeypairFromFileSync. Removed the helper and imports from wallet-loader.ts.

Tests updated: expandPath describe block (5 tests) ported to wallet.test.ts + new loadKeypairFromFile tilde integration test. wallet-loader.test.ts replaced with module smoke tests confirming expandPath is no longer exported from that module.

- wallet.ts targeted: 30/30 ✅ (+7 new)
- wallet-loader.ts targeted: 2/2 ✅
- Full suite: 17 pre-existing failures, 0 regressions

PR #456, opened 2026-04-18
fix(tools): rename to taskDescription to avoid JSON Schema collision

### LINKED ISSUE #455: bug(tools): taskDescription field name collides with JSON Schema reserved keyword causing silent tool failures
## Description

The `taskDescription` field name in the AgenC tool definitions in `runtime/src/tools/agenc/tools.ts` collides with the JSON Schema `description` keyword when the tool schema is serialized and passed to LLM providers. This causes silent failures where the tool input is not correctly parsed or the field is dropped entirely by the schema validator.

## Affected file

`runtime/src/tools/agenc/tools.ts`

## Root cause

JSON Schema uses `description` as a reserved keyword at the schema level. When a tool input field is named `taskDescription`, some schema validators and LLM provider SDKs confuse it with the schema-level `description` property, causing the field to be silently ignored or misrouted.

## Fix

Rename `taskDescription` to a non-colliding name (e.g. `task_description` or `jobDescription`) throughout `tools.ts` and any callers. The rename should be consistent across the tool definition, input validation, and any downstream consumers of the field.

## Impact

Silent tool failures when creating or referencing tasks via the AgenC tool surface. Affects any agent workflow that passes task descriptions through the tool registry.

letterj:
Closes #455

## Summary

The `create_task` tool in `runtime/src/tools/agenc/tools.ts` used `description` as the input field name for the task description. This collides with the JSON Schema `description` keyword at the schema level, causing some LLM provider SDKs and schema validators to silently ignore or misroute the field.

## Changes

**`runtime/src/tools/agenc/tools.ts`** (+4 lines, -4 lines):
- Schema property renamed: `description` → `taskDescription`
- - `required` array updated: `'description'` → `'taskDescription'`
- - Dedup key updated with backwards-compat fallback: `args.taskDescription ?? args.description ?? ''`
- - `parseTaskDescription` call updated with backwards-compat fallback: `args.taskDescription ?? args.description`
**`runtime/src/tools/agenc/tools-task-templates.test.ts`** (+26 lines):
- 2 new tests asserting `taskDescription` is present and `description` is absent in both schema properties and required array
## Test results

- 35/35 targeted tests passing (src/tools/agenc/)
- - TypeScript: 0 errors

### DIFF
--- a/runtime/src/tools/agenc/tools-task-templates.test.ts
+++ b/runtime/src/tools/agenc/tools-task-templates.test.ts
@@ -110,7 +110,7 @@ describe("agenc task template tools", () => {
     });
 
     const result = await tool.execute({
-      description: "Devnet ABI warning task",
+      taskDescription: "Devnet ABI warning task",
       reward: "1",
       requiredCapabilities: "1",
       taskId: "11".repeat(32),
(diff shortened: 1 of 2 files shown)


7etsuo:
Merged — thanks, the backwards-compat fallback on the read sites is nice.

PR #458, opened 2026-04-18
fix(tools): filter for active agent status in resolveAuthorityAgentPda

### LINKED ISSUE #457: bug(tools): resolveAuthorityAgentPda does not filter for active agents, can resolve to suspended or deregistered agent PDA
## Description

`resolveAuthorityAgentPda` in `runtime/src/tools/agenc/mutation-tools.ts` resolves the agent PDA for the current signer wallet without filtering for active agent status. If an agent has been suspended or deregistered, the function still resolves and returns the stale PDA, causing downstream tool calls to fail with on-chain instruction errors that are hard to diagnose.

## Affected file

`runtime/src/tools/agenc/mutation-tools.ts`

## Root cause

The PDA resolution logic finds the agent account for the signer wallet but does not check the `status` field of `AgentRegistration`. A suspended (`status: suspended`) or deregistered agent PDA will be returned, and subsequent on-chain instructions that require an active agent will fail with opaque errors.

## Fix

Add an active-status check in `resolveAuthorityAgentPda` — after resolving the PDA, verify that `AgentRegistration.status === 'active'` before returning. If the agent is not active, return a descriptive error indicating the agent is suspended or deregistered rather than silently returning a stale PDA.

## Impact

Any agent tool call that requires an active agent PDA can silently use a stale PDA, producing hard-to-diagnose on-chain errors. Affects task creation, claiming, completion, and other mutation operations routed through this function.

letterj:
Closes #457

## Summary

`resolveAuthorityAgentPda` in `mutation-tools.ts` previously returned the first matching agent PDA for the signer wallet without checking agent status. A suspended or deregistered agent PDA could be silently returned, causing opaque on-chain errors in downstream tool calls.

The fix fetches full account data for all candidate PDAs, filters to `AgentStatus.Active` only, auto-selects if exactly one active agent remains, and surfaces active-only candidates in the ambiguity error when multiple active agents exist.

## Changes

**`runtime/src/tools/agenc/mutation-tools.ts`** (+2 lines, -2 lines):
- Added `AgentStatus` to import from `../../agent/types.js`
- - Replaced immediate ambiguous-signer return with active-status disambiguation: fetches full account data via `fetchMultiple`, filters to `AgentStatus.Active`, auto-selects if exactly one active remains, otherwise passes active-only candidates to the ambiguity error
**`runtime/src/tools/agenc/mutation-tools.test.ts`** (+53 lines):
- Added `fetchMultiple: vi.fn(async () => [])` to base mock — safe default preserving all existing tests
- - New test: single active agent auto-selected when one of two registrations is suspended
- - New test: ambiguity error shows only active PDAs when 2 active + 1 suspended are present
## Functional testing

**Tested on devnet (V3 program `2jdBSJ8U5ixfwgs1bRLPtRRnpZAPm8Xv1tEdu8yjHJC7`):**
- ✅ Happy path — active creator agent resolves correctly, task created successfully (task PDA: `3t7pyGyHa6dkBMseZ715cref52mH5P83fyV5P3RJHXTs`)
**Not functionally testable without protocol authority:**
- ❌ Suspended agent path — requires `suspend_agent` instruction (protocol authority only)
- - ❌ Deregistered agent path — requires controlled deregister + slash window management
These paths are covered by unit tests only.

## Test results

- 35/35 targeted tests passing (src/tools/agenc/)
- - TypeScript: 0 errors

### DIFF
--- a/runtime/src/tools/agenc/mutation-tools.test.ts
+++ b/runtime/src/tools/agenc/mutation-tools.test.ts
@@ -107,6 +107,9 @@ function createMockProgram(options: { signer?: PublicKey | null } = {}) {
           if (pda.equals(DEFENDANT_AGENT_PDA)) return makeRawAgent(DEFENDANT_WALLET);
           throw new Error(`Unknown agent: ${pda.toBase58()}`);
         }),
+        // Default: return empty array so existing tests are unaffected.
+        // Override per-test to exercise the active-status filter path.
+        fetchMultiple: vi.fn(async () => []),
       },
       skillRegistration: {
         fetch: vi.fn(async (pda: PublicKey) => {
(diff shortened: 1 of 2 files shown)


7etsuo:
Merged — thanks for the active-status filter and the regression tests.

PR #1628, opened 2026-07-31
docs: sync daemon method count in sdk.md with protocol constant

latent-9:
The daemon method surface in `docs/sdk.md` is out of sync with the authoritative constant.

`AGENC_SDK_DAEMON_METHODS` in `packages/agenc-sdk/src/protocol.ts` defines **48** methods, but the docs say:
- line 77 (prose): "any of the **46** daemon methods"
- line 256 (heading): "Daemon method surface (46 methods)"
- the surface table (lines 263-272) lists only **47** — it omits `session.resolveToolCall`.

`session.resolveToolCall` is a real method: it is in the constant (`protocol.ts:53`) and has typed params/result (`SessionResolveToolCallParams` at `protocol.ts:467`, `SessionResolveToolCallResult` at `protocol.ts:1197`).

This PR:
- updates both count references `46` → `48`, and
- adds `session.resolveToolCall` to the `sessions` group (in the same order as the constant, between `session.cancelTurn` and `session.mcp.addServer`).

After this, prose, heading, table, and the `protocol.ts` constant all agree on 48. Docs-only, no functional change.

### DIFF
--- a/docs/sdk.md
+++ b/docs/sdk.md
@@ -74,7 +74,7 @@ Key `AgencClient` methods:
   `{ runId, specDigest, baseCommit, baseDirty }`)
 - `runStatus(id)` / `runResult(id)` / `replayRun(params)` /
   `reattachRun(options)` / `runEvidence(params)` / `cancelRun(id, reason?)`
-- `request(method, params)` → raw typed JSON-RPC for any of the **46** daemon methods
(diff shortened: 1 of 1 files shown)


latent-9:
Closing this: the daemon method count on main is already correct (51), and this PR's value (48) does not match the AGENC_SDK_DAEMON_METHODS constant (which enumerates 51). The branch is also out of date and conflicts. Apologies for the noise.

PR #1699, opened 2026-08-11
fix(tui): keep tool output minimal and mark every in-flight label with one star

signerless:
Two reports from a live session, both about the transcript being noisier and
flatter than it needs to be.

### The whole fetched page in the chat

WebFetch printed every fetched document in full. Only tools that declare
`isSearchOrReadCommand` are folded into the collapsed group summary the way
`Read` and `Grep` are, and neither web tool declared it — a fetch IS a read and
a search IS a search. Both now do.

Any tool with no renderer of its own also gets a floor: past 2k characters the
transcript keeps the first 12 lines and says how many it hid, pointing at
ctrl+o for the rest. The floor is not the mechanism — a tool whose output is
routinely long should declare itself — but nothing can bury a screen of
conversation again.

### An animated glyph for a running tool

The mark beside a tool call was a lifecycle circle (`○ ◐ ● ✕`). A running tool
now shows a star whose points sweep.

The unicode frames come from the star family already proven in this TUI: a bare
ASCII `*` among unicode stars renders thin and raised and visibly flickers each
cycle, which is why `getDefaultCharacters` avoids it too. ASCII mode rotates the
arms instead — `+` axis-aligned, `*` six-armed, `x` diagonal — which is the same
idea with characters every terminal has.

Only the `running` state animates, so a transcript full of finished rows stays
off the animation clock, and reduced motion freezes on a glyph distinct from
`done` (otherwise a user with animations off could not tell a live tool from a
finished one — the one thing that column exists to say). It ticks slower than
the composer spinner: several rows can be in flight at once and a fast sweep
down a column reads as noise.

Scoped to the tool-call path only. Plan glyphs, the header tab dot and the
background-tasks panel are untouched.

### Verification

Existing assertions that pinned the old circles were updated, including the
`v2-primitives` shape tests — those are about the glyph/space/label/space/paren
pinning contract, not about which glyph it is, so they now bind to the real
glyph and the shape stays pinned while the mark is free to change.

TUI components, message renderers, coverage swarm and v2 primitives: 1268
passing.

### DIFF
--- a/runtime/src/tools/WebFetchTool/WebFetchTool.ts
+++ b/runtime/src/tools/WebFetchTool/WebFetchTool.ts
@@ -117,6 +117,14 @@ export const WebFetchTool = buildTool({
   // tool for the rest of the run. Observed twice in one afternoon, each time
   // stranding a live hardware-debugging session that could no longer flash.
   recoveryCategory: 'idempotent',
+  // A fetch IS a read, and its result is a whole web page. Without this the
+  // transcript printed every fetched document in full (report: "the webchat is
+  // showing the complete fetch"), because only tools that declare themselves
+  // search/read get folded into the collapsed group summary the way Read and
+  // Grep are.
+  isSearchOrReadCommand() {
+    return { isSearch: false, isRead: true }
+  },
   toAutoClassifierInput(input) {
     return input.prompt ? `${input.url}: ${input.prompt}` : input.url
   },
--- a/runtime/src/tools/WebSearchTool/WebSearchTool.ts
+++ b/runtime/src/tools/WebSearchTool/WebSearchTool.ts
@@ -405,6 +405,11 @@ export const WebSearchTool = buildTool({
   // poison the session. See the note in WebFetchTool for what the
   // side-effecting default costs.
   recoveryCategory: 'idempotent',
+  // A search is a search: fold the result list into the collapsed group
+  // summary instead of printing every hit, matching Grep and Glob.
+  isSearchOrReadCommand() {
+    return { isSearch: true, isRead: false }
+  },
   toAutoClassifierInput(input) {
     return input.query
   },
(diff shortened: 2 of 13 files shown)


signerless:
Rebased onto main (now past #1698 and #1700) and extended: the running star introduced here now carries every in-flight label in the transcript — permission waits, auto/bash classifier checks, and collab rows that are still running (`Waiting for agents`, `Spawned …`). Settled collab rows keep their static mark, so the star still means exactly one thing.

Note on the earlier red `neovim-darwin-arm64`: three reruns failed on three *different* tests, every time with `Embedded Neovim startup timed out after 20000ms while attaching the embedded UI` — a slow macOS ARM runner, not a logic fault, which would have failed the same test each time. The same job failed repeatedly on #1697 before that merged clean. Nothing in this diff touches Neovim, PTY lifecycle, or the failing scenario.

signerless:
Merging with `neovim-darwin-x64` red. Evidence it is the runner and not this change:

- **4 failures across 4 reruns, on 4 different tests**, on **both** macOS runners (arm64 and x64). Every one with the same message: `Embedded Neovim startup timed out after 20000ms while attaching the embedded UI`. A logic fault would fail the same test each time.
- The failing file is `buffer-neovim-lifecycle.real-neovim.test.ts` and one PTY scenario (`131-workbench-buffer-neovim-platform-kill-cleanup`). This diff touches the tool glyph, the transcript result clamp, and two web-tool flags — no Neovim, no PTY, no buffer lifecycle. Neither the file nor the scenario references the glyph.
- The same macOS jobs failed repeatedly on #1697 (`neovim-darwin-arm64`, `neovim-darwin-x64`, `macos-native`) before that branch merged clean.
- All 8 other checks pass, including `linux-kernel-sandbox`, `macos-native` and `windows-native`.

Locally on this branch: typecheck clean, `tests/tui/message-renderers` + `tests/tui/components` 1247 passing.

PR #1701, opened 2026-08-11
feat(tui): replace the agent scanline with a dot bar that fills with work

signerless:
The running-agent row drew a marker walking back and forth along a line:

```
› Cyberia · Scanner
  running · 3m24s
  ·······—◇·············  14 tools · 355.4k tok
```

It said "alive" and nothing else, and it reads as a figure pacing rather than as progress.

It is now a bar of dots that fills left to right, one dot per tool call, grey to bright.

### Why dots and not a percentage

What it measures is **work done, not percent complete**. AgenC gets no trustworthy completion estimate from a running agent, so a dot per real tool call is a fact where a `58%` fill would be an invention. That honesty was the whole reason for the scanline; this keeps it and adds real information, since the driving count is the same one already printed beside the bar.

Past a full bar it wraps, and the remainder brightens one step so a second pass is visibly a second pass rather than looking like the agent started over — the `N tools` counter beside it makes the lap unambiguous. The next dot pulses slowly so the row still reads as alive between tool calls; reduced motion freezes it.

### Verification

The bar keeps exactly `width` cells at every fill level. A test for that caught a real defect while writing it: a non-finite tool count poisoned every `repeat()` into `""` and collapsed the bar to zero cells — an invisible row rather than an empty one. Guarded in the source, not weakened in the test.

`tests/tui/workbench/agents`: 36 passing. The rail's own assertion moved from the walking marker to the bar.

### DIFF
--- a/runtime/src/tui/workbench/agents/AgentActivityTrack.tsx
+++ b/runtime/src/tui/workbench/agents/AgentActivityTrack.tsx
@@ -2,77 +2,89 @@ import React, { useEffect, useState } from "react";
 
 import { Text } from "../../ink.js";
 
-const ACTIVITY_TRACK_FRAME_MS = 80;
+const ACTIVITY_PULSE_FRAME_MS = 450;
 const MIN_TRACK_WIDTH = 6;
 
 export type AgentActivityTrackFrame = {
-  readonly after: string;
-  readonly before: string;
-  readonly head: "◆" | "◇";
-  readonly tail: string;
+  /** Dots for work already done, drawn bright. */
+  readonly filled: string;
+  /** The dot about to be earned; pulses so the row reads as alive. */
+  readonly head: string;
+  /** Work not done yet, drawn grey. */
+  readonly rest: string;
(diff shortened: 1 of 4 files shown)


signerless:
Merging with the two macOS jobs red. Both are deadline timeouts on runners that are degraded today, not results:

- `neovim-darwin-arm64`: `Embedded Neovim startup timed out after 20000ms while attaching the embedded UI` — the fifth such failure this session, on the fifth *different* test, across both macOS runners.
- `macos-native`: `inventory untracked production files failed or exceeded its 5000 ms deadline` / `bounded metadata command stopped for timeout`.

This diff touches one file in the agents rail plus its tests. It has no relationship to embedded Neovim, PTY lifecycle, or the FND inventory harness. Both jobs also failed repeatedly on #1697 before that branch merged clean, and all 7 other checks pass here — including `linux-kernel-sandbox` and `windows-native`.

Locally: typecheck clean, `tests/tui/workbench/agents` 36 passing.

PR #1704, opened 2026-08-13
feat(models): catalogue grok-4.6 and make it the startup default

signerless:
xAI shipped `grok-4.6` on 2026-08-12. It was missing from `REGISTERED_MODEL_CATALOG`, so it never appeared in `/model` and callers fell back to name heuristics for its context window.

Specs pinned against [docs.x.ai/developers/grok-4-6](https://docs.x.ai/developers/grok-4-6): 500k context, text + image input, tools and structured output, reasoning `low/medium/high/xhigh` defaulting to `high`.

### The one thing that could not reuse an existing constant

`xhigh`. Every earlier Grok stops at `high`, so reusing `GROK_REASONING_LEVELS` would have silently dropped the level xAI documents and left the top effort tier unreachable from the picker. 4.6 gets its own tuple, and a test asserts 4.5 still does *not* carry `xhigh` so the two never drift back together.

### What this does not change

The **default stays `grok-4.5`**. 4.6 leads the grok list as the newest model, but listing a model and defaulting to it are separate decisions — the default is hardcoded in six places (`config/env.ts`, `gateway/x-search.ts`, `bin/startup-selection.ts`, `utils/model/model.ts`, `heartbeat/runner.ts`, and the `agenc-main` docs) and moving it is its own change with its own cost implications.

### Verification

Five new assertions cover: resolves as `visibility: "list"` (`hide` would keep it resolvable but drop it from the picker — the exact symptom), 500k resolved *through the catalog* rather than a heuristic, the `xhigh` tuple, modalities/tools/structured-output including the existing `grok-4` family gate, and ordering ahead of 4.5.

Two existing tests pinned 4.5 as first in the picker and now pin the new order; both keep their default-model assertions intact.

`tests/llm`, `tests/config`, `tests/commands`: 1804 passing. Typecheck clean.

Probed against the live xAI API, not just the catalog:

```
$ AGENC_MODEL=grok-4.6 agenc -p "Reply with exactly: GROK46_OK"
GROK46_OK
```

### DIFF
--- a/runtime/src/bin/startup-selection.ts
+++ b/runtime/src/bin/startup-selection.ts
@@ -16,7 +16,7 @@ import type { AgenCConfig } from "../config/schema.js";
 import { tokenizeCliOptionRegion } from "./cli-option-region.js";
 import { extractFlagValue } from "./route.js";
 
-const DEFAULT_MODEL = "grok-4.5";
+const DEFAULT_MODEL = "grok-4.6";
 
 export const PROVIDER_MODEL_CATALOG = BUILT_IN_PROVIDER_MODEL_CATALOG;
 
--- a/runtime/src/config/env.ts
+++ b/runtime/src/config/env.ts
@@ -214,7 +214,7 @@ export function resolveProviderBaseURL(
 
 /** Model slug from env, falling back to `defaultModel`. */
 export function resolveModel(
-  defaultModel = "grok-4.5",
+  defaultModel = "grok-4.6",
   env: EnvSnapshot = process.env,
 ): string {
   const e = readEnv(env);
--- a/runtime/src/config/schema.ts
+++ b/runtime/src/config/schema.ts
@@ -946,7 +946,7 @@ export const KNOWN_CONFIG_KEYS: readonly string[] = Object.freeze([
 export function defaultConfig(): AgenCConfig {
   return Object.freeze({
     configVersion: 1,
-    model: "grok-4.5",
+    model: "grok-4.6",
     model_provider: "grok",
     approval_policy: "on-request" as ApprovalPolicy,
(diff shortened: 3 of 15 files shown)


signerless:
Extended: **grok-4.6 is now the startup default**, not just catalogued.

The default is written in five places and all five move together — the provider's built-in default, the env resolution default, the shipped `defaultConfig`, startup selection, and the heartbeat admit fallback.

Two things had to move with it, and neither is cosmetic:

**Vision + tools.** `VISION_MODELS_WITH_TOOLS` in the grok adapter gates which vision models may carry client-side function tools. 4.6 takes images *and* supports tools, but was not in that set — promoting it to default without this entry would have quietly dropped function calling on any turn containing an image.

**Pricing.** `DEFAULT_MODEL_COSTS` feeds `dollar_cap` enforcement, so an unpriced default bills at `DEFAULT_UNKNOWN_MODEL_COST` and enforces budgets at the wrong threshold. xAI bills 4.6 exactly as 4.5 below 200k prompt tokens ($2 in / $0.50 cached / $6 out per 1M), so it reuses that tier.

> **Known limitation:** above 200k prompt tokens xAI doubles the rate ($4 / $1 / $12) and `ModelCostEntry` has no prompt-size tier, so a >200k turn is under-counted. Under-counting is the deliberate side to err on — the alternative trips `dollar_cap` early on every short turn. Adding tiered pricing is its own change.

The openrouter default (`x-ai/grok-4.5`) is a different provider's default and is left alone.

Six existing tests pinned 4.5 as the default and now pin 4.6; the ones that set an explicit `model = "grok-4.5"` and assert it round-trips were left untouched, since those test that an explicit value is honoured.

`tests/llm`, `tests/config`, `tests/commands`, `tests/session`: **2901 passing**. Typecheck clean.

PR #1728, opened 2026-08-15
fix(process): let a closing descendant settle before calling it residue

signerless:
## The flake, reproduced

`bounded-temp-repository` and friends have been failing in CI with a commit that **succeeded**:

```
BoundedRepositoryGitError: failed to commit repository files (exit=0, signal=null, stop=residual_process)
```

It does not reproduce on an idle machine — 0 of 6. Under 24 busy processes it reproduces on demand:

```
fallos: 5/6
```

That reproduction is the point of this PR. A previous attempt at this bug was withdrawn because it was iterated blind against CI and broke macOS twice.

## Cause

A leader's `close` fires once its own stdio is done and it has been reaped. That says nothing about a helper it spawned moments earlier, which can still sit in the process table while it tears itself down. The classifier looked **once, at that instant**:

```js
} else if (isProcessTreeAlive(child)) {
  requestStop("residual_process");
}
```

Under contention that gap widens enough to be observed routinely, so git exits 0 and the commit is reported as failed anyway.

## Fix

A descendant seen alive at close gets a bounded window (250 ms, early exit as soon as the tree goes quiet) to finish exiting before it is classified.

The discrimination is exact, not heuristic:

| | at close | after the window |
|---|---|---|
| Genuine leak (contract test: detached, ignores SIGTERM, `setInterval` forever) | alive | **still alive** → `residual_process`, unchanged |
| Helper on its way out (this flake) | alive | gone → clean exit |

The wait is paid only when something is alive at close at all, which is rare.

## What the earlier attempt got wrong

Both deliberate here:

- The subreaper's **authoritative** `residual` flag still decides immediately, without observing again. It is a decision, not an observation.
- `close` no longer reaches `maybeFinish` before classification. Finishing early is exactly what let a real leak report a clean exit in the first withdrawn version.

## Verification

| | |
|---|---|
| Under the load that produced 5/6 failures | **0 of 8** |
| `supervisedProcess` contract | 29 passed |
| `red-probe-runner` contract (asserts `stop=residual_process`) | 67 passed |
| process/temp-repo family | 98 passed |
| typecheck | clean |

macOS and Windows are covered by CI here, not by local runs — that is where the withdrawn attempt failed, so it is the part worth watching on this PR.

### DIFF
--- a/runtime/src/utils/supervisedProcess.ts
+++ b/runtime/src/utils/supervisedProcess.ts
@@ -87,17 +87,23 @@ const DEFAULT_TERMINATE_GRACE_MS = 500;
 const DEFAULT_SETTLE_BACKSTOP_MS = 1_000;
 const PROCESS_TREE_POLL_INTERVAL_MS = 20;
 /**
- * How long a descendant still visible when the leader closes is given to finish
- * exiting before it counts as residue.
+ * How long a Linux-cgroup member still visible when the leader closes is given
+ * to finish exiting before it counts as residue.
  *
  * The leader's `close` fires once its own stdio is done and it has been reaped,
  * which says nothing about a helper it spawned moments earlier: that helper can
- * still be in the process table while it tears itself down. On a loaded machine
+ * still populate the cgroup while it tears itself down. On a loaded machine
  * that gap widens enough to be observed routinely, so an instantaneous look
- * reports residue for a tree that is already on its way out. A real leak -- a
- * detached descendant that outlives its leader -- is still there when this
- * window closes, so the distinction costs nothing but a bounded wait, and only
- * in the rare case where anything is alive at close at all.
+ * reports residue for a tree that is already on its way out. A real leak is
+ * still inside the cgroup when this window closes -- containment guarantees it
+ * cannot slip away unseen -- so the distinction costs nothing but a bounded
+ * wait, paid only when anything is alive at close at all.
+ *
+ * This window applies ONLY to the cgroup boundary. The other containment
+ * paths either report residue authoritatively (Linux subreaper, Windows job)
+ * or can observe it only in the instant of close (the darwin POSIX gate
+ * settles its own process group during teardown), so they classify
+ * synchronously; waiting there erases the signal instead of de-noising it.
  */
 const RESIDUAL_SETTLE_WINDOW_MS = 250;
 const PROCESS_TABLE_TIMEOUT_MS = 2_000;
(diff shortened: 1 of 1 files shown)


signerless:
Marking this a draft: **the approach is unsound on macOS** and CI caught it.

`macos-native` fails the contract test that exists for exactly this:

```
FND red-probe supervisor > keeps the standalone audit alive until a residual descendant is settled
AssertionError: expected +0 to be 1
```

The audit now exits 0 where it must exit 1 with `stop=residual_process` — a genuine leak reported as a clean exit, which is the same failure mode as the earlier withdrawn attempt, reached from the other direction.

## Why

The settle window assumes process-tree liveness is stable over time. It is not, on the darwin path.

With no subreaper and no cgroups, `isProcessTreeAlive` re-derives membership on every call:

```js
if (!linuxCgroupBoundaries.has(child)) {
  captureProcessTreeDescendants(child);
}
const ownedDescendantAlive = ownedBoundaryHasLiveMember(child);
```

`captureProcessTreeDescendants` walks parentage from `rootPid`. Once the leader exits, a detached descendant is reparented to launchd and is no longer a descendant of that root, so the re-check sees a quiet tree. The process did not exit — it stopped being visible. Waiting and looking again therefore cannot distinguish "settled" from "reparented" on macOS, which is the whole premise of this change.

On Linux the boundary keeps its own membership, so the premise holds there: 5/6 failures under load became 0/8.

## Path forward

Re-check the **specific descendants observed at close**, by pid plus the identity the boundary already tracks (`boundary.identities`, which exists to survive pid reuse), instead of re-deriving the tree. A reparented process keeps its pid and identity, so it stays visible for the whole window on every platform.

Leaving the reproduction recipe here since it is the part worth keeping: the flake does not appear on an idle machine (0/6), and reproduces on demand under ~24 busy processes (5/6).

signerless:
Correction to my previous comment: the reparenting diagnosis was wrong. `boundary.identities` is cumulative, so a reparented descendant that was ever captured stays visible — reparenting was never the mechanism.

The actual bug was in the patch itself: the settle timer was **unref'd**. After `close`, a standalone runner can have an otherwise-empty event loop, so the process exited with no verdict at all — the audit's 0 was "never classified", not "misclassified". Reproduced on Linux with a standalone probe (unresolved top-level await, rc=13, in both the linger and settle scenarios), no Mac required.

Running the same probe against **main** shows both scenarios also die without a verdict there — the runner-exits-before-verdict hazard is pre-existing for every deferred path (`treeExitTimer`, `forceTimer`, `backstopTimer` are all unref'd). The broken commit widened it to close-time classification, which used to be synchronous; the fix narrows it back and goes one better: the settling helper now gets a clean verdict where main produced none.

New commit:
- settle timer referenced (bounded by the deadline; at expiry it hands off to the exact `requestStop("residual_process")` main ran synchronously at close)
- clean finish requires two consecutive quiet polls, so one failed `/bin/ps` snapshot on darwin (which wipes `boundary.current`) cannot turn a live descendant into a clean exit

Local: probe verdicts correct on both scenarios, 0/8 under the load that reproduced 5/6, supervisedProcess 29 + red-probe 67 + temp-repo/process family (165 total), typecheck clean. `macos-native` remains the judge here.

signerless:
Third diagnosis, this time with the darwin path executed rather than reasoned about. `macos-native` failed again on the referenced-timer commit, so the unref was necessary but not sufficient.

**Reproduction without a Mac**: spoofing `process.platform = "darwin"` before module load runs the real darwin branch on Linux — the detached POSIX gate, process groups, `kill(-pid, 0)`, `/bin/ps -axo` all execute natively. With the exact red-probe fixture shape (in-group descendant sleeping 4s, leader exits after 250ms):

| | verdict | descendant after |
|---|---|---|
| main | `residual_process` | settled |
| previous commit | `undefined` (clean) | **also dead** |

The second row is the tell: nobody classified the residue, yet the descendant died anyway. **The darwin POSIX gate settles its own process group during teardown**, so residue is observable only in the instant of `close`. A settle window there waits out the *gate's cleanup*, then reads the swept tree as clean — it erases the signal instead of de-noising it.

The Linux cgroup boundary is the opposite case: authoritative containment, so "populated at close, empty moments later" genuinely means the helper finished exiting, and a real leak cannot leave the cgroup unseen. That is also the boundary the flaking CI family (default-suite) runs under.

**The fix**: the settle window now applies only when a cgroup boundary exists. Subreaper, darwin gate, and Windows job paths classify synchronously at close — byte-for-byte main behavior.

Validation: darwin-path probe matches main (`residual_process`, descendant settled); cgroup probe gives the settling helper a clean verdict where main produced none; 0/8 under the load that reproduced 5/6; supervisedProcess 29 + red-probe 67 + temp-repo/process family (165 total); typecheck clean.

PR #1733, opened 2026-08-16
feat(sandbox): vendor the Landlock launcher and surface its enforcement in doctor

signerless:
Adoption #2 from the deepseek-harness study. **Foundation slice, deliberately report-only**: vendor the launcher, prove it end to end on a real kernel, and tell the user what their machine can enforce. Wiring it into the execution path is a separate, deliberate change.

## Why Landlock

It is an independent kernel syscall family — no user namespaces, no mount privileges, no LSM profile. That is exactly the gap our sandbox has today: on Ubuntu 24.04+ with `apparmor_restrict_unprivileged_userns=1`, bubblewrap is dead on arrival and the only remediation is a sudo-installed AppArmor profile. Landlock enforces without any of that. This host demonstrates the case:

```
  Sandbox:            unavailable (mode: workspace_write, platform: linux)
    reason:   Linux sandbox helper must be outside the writable workspace
    landlock: fully enforced          ← new
```

## What's in the box

- **`runtime/native/agenc-landlock-run.c`** — vendored verbatim from DeepSeek Harness's `landlock-run` (MIT, commit `47f943859`), ~280 lines of C11 over the raw Landlock UAPI, self-defining the ABI so the build is independent of header vintage. Self-restrict-then-exec: installs the ruleset on itself, execs the command; the ruleset is inherited across `execve`, so every descendant is equally confined. Fail-closed when the kernel cannot enforce. The CLI contract (`--ro/--rw` grants, `--probe`, the `landlock-run: ` prefix, exit 125) is preserved **verbatim** so upstream fixes stay diffable and the calibrated classification below holds.
- **Build**: mirrors the process broker exactly — compiled to `dist/` by `build.config.ts` with the same hardened flags, listed in `agencExecutableFiles`, compiled on demand from source with the trusted-compiler pattern (fixed absolute compiler paths, 0700 build dir) when no bundled binary exists.
- **`src/sandbox/landlock-run.ts`** — resolution across both bundle geometries, functional probe (`full`/`partial`/`unusable`; actually restricting a short-lived process is the only honest signal — a version check would miss a kernel that has the syscalls but refuses enforcement), grant serialization, and outcome classification.
- **Classification is postmortem-calibrated**: upstream's postmortem 0004 documents how the partial-ABI notice plus ripgrep's no-match exit 1 got misread as sandbox failure. Launcher failure here requires the conjunction — exit **125** AND a `landlock-run: ` fatal line — after excluding the **exact** informational notice. Falsification-checked: removing the exclusion fails exactly that test.

## Live verification (kernel ABI 8)

| Semantics | |
|---|---|
| Read outside grants denied / inside allowed | ✅ |
| Write denied under `--ro`, allowed under `--rw` | ✅ |
| Descendants inherit the ruleset | ✅ |
| Child exits pass through (7→7, `false`→1, no launcher lines) | ✅ |
| Unopenable grant root fails closed at 125 | ✅ |
| Built `agenc doctor` prints the enforcement line | ✅ |

12 landlock tests (live suite self-skips without linux + trusted cc + enforceable kernel, per the pwsh precedent), supervisedProcess 29, doctor suites 27, typecheck and build clean. `REQUIRED_GATE_RED_PROBE_POLICY_PATHS` untouched — no gate-policy rotation needed.

## Next (separate PRs)

1. Execution-path fallback: when bwrap is unavailable and the probe reports `full`, offer Landlock confinement for `PermissionProfile` mounts (path grants map directly; network stays seccomp's job).
2. Installer note: replace the bare sudo remediation with "or rely on Landlock" once the fallback lands.

### DIFF
--- a/runtime/build.config.ts
+++ b/runtime/build.config.ts
@@ -45,6 +45,14 @@ const processBrokerDist = resolve(
   runtimeRoot,
   'dist/agenc-process-broker',
 );
+const landlockRunSource = resolve(
+  runtimeRoot,
+  'native/agenc-landlock-run.c',
+);
+const landlockRunDist = resolve(
+  runtimeRoot,
+  'dist/agenc-landlock-run',
+);
 const windowsProcessBrokerSource = resolve(
   runtimeRoot,
   'native/agenc-process-job-broker.cs',
@@ -163,6 +171,44 @@ function compileLinuxProcessBroker(): void {
     );
   }
   chmodSync(processBrokerDist, 0o755);
+}
+
+function compileLinuxLandlockRun(): void {
+  if (process.platform !== 'linux') return;
+  const compiler = process.env.CC?.trim() || 'cc';
+  const result = spawnSync(
+    compiler,
+    [
+      '-O2',
+      '-std=c11',
+      '-Wall',
+      '-Wextra',
+      '-Werror',
+      '-D_FORTIFY_SOURCE=2',
+      '-fstack-protector-strong',
+      '-Wl,-z,relro,-z,now',
+      '-o',
+      landlockRunDist,
+      landlockRunSource,
+    ],
+    {
+      cwd: runtimeRoot,
+      env: {
+        ...process.env,
+        LANG: 'C',
+        LC_ALL: 'C',
+      },
+      encoding: 'utf8',
+    },
+  );
+  if (result.error !== undefined || result.status !== 0) {
+    throw new Error(
+      'Linux landlock-run build failed' +
+        (result.error === undefined ? '' : `: ${result.error.message}`) +
+        (result.stderr ? `\n${result.stderr.trim()}` : ''),
+    );
+  }
+  chmodSync(landlockRunDist, 0o755);
 }
 
 function compileWindowsProcessBroker(): void {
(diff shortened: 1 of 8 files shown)


signerless:
Design note for the execution-path fallback (the announced follow-up), recorded before starting it: **the crux is network confinement, and it forks the design.**

Under bubblewrap, AgenC's network story is `--unshare-net` plus a seccomp BPF program that **bwrap itself applies** (`--seccomp <fd>`; the program is built in `linux-launcher/landlock.ts` and handed over via `SECCOMP_STDIN_FD`). The vendored Landlock launcher confines the **filesystem only** — the fs allow-list — and applies no seccomp. A naive fallback (bwrap unavailable → wrap with `agenc-landlock-run`) would therefore ship a sandbox that silently stops confining the network on exactly the hosts that fall back. That is the silent-degradation class this codebase refuses everywhere else.

Three honest options:

1. **Extend the vendored launcher** to also install a seccomp program read from an inherited fd (mirroring bwrap's `--seccomp` contract). Keeps one self-restrict-then-exec binary; diverges from the upstream CLI contract we deliberately preserved verbatim.
2. **Landlock ABI-4 TCP rules** (`bind_tcp`/`connect_tcp`) in the launcher: kernel-level network allow-listing with no seccomp at all. Cleaner long-term; more UAPI surface to vendor and negotiate, and it governs TCP only.
3. **Fs-only fallback, loudly labeled**: fall back for filesystem confinement and surface `network: unconfined (landlock fallback)` in the status, doctor, and the permission prompt. Simplest and honest, but weaker than what bwrap users get.

None of these is obviously the maintainers' call to skip. Foundation stays report-only until this is decided.

PR #1762, opened 2026-08-21
fix(workflow): derive child agent names the registry accepts

signerless:
## What breaks

Every verified-change run dies at `workflow.plan`, before any work happens:

```
$ agenc run start --goal "add a file called probe.txt containing ok" \
    --verify "check=test -f probe.txt"

run wf-da1faa33-… — failed (terminal)
STAGE                 STATUS       ATTEMPTS
workflow.intake       committed    1
workflow.worktree     committed    1
workflow.plan         failed       2
workflow.implement    blocked      0
workflow.verify       blocked      0
workflow.review       blocked      0
workflow.finalize     blocked      0
stop reason: step_retries_exhausted
```

The daemon records why in `run_terminal_results`:

```
workflow plan spawn rejected: agent_name must use only lowercase letters,
digits, and underscores
workflow.plan failed terminally after 2 attempt(s)
```

## Why

The two ends of the same name disagree.

`session-adapters.ts` derived the child's agent name from its child run id
and kept dashes, dots and uppercase:

```ts
agentName: input.childRunId.replace(/[^a-zA-Z0-9._-]/g, "-")
```

`assertValidAgentName` in `agents/registry.ts` accepts none of those:

```ts
if (!/^[a-z0-9_]+$/u.test(name)) throw new InvalidAgentPathError(...)
```

A child run id looks like `wf-3f78249a-c5e4-42b4-90ac-c89cf87618f5:plan#1`,
so the derived name kept every dash and was rejected on arrival.

## The change

Fold the separators to underscores, behind a named helper so the rule lives
next to the id it transforms, and add a test that runs the derived names
through the registry's own validator rather than restating the pattern.

## How it was checked

Reproduced on 0.16.1 through both `agenc run start` and the daemon RPC, in
three different repositories, with and without an explicit `--model`, so
neither the model string nor the repository name was involved.

With the same change applied to the installed runtime, the rejection is
gone and the plan child spawns: `workflow.plan` no longer reports
`spawn rejected`. Those runs still end at plan on this machine, but for an
unrelated reason — the provider route is returning `HTTP 500
provider_unavailable`, which fails plain sessions too.

New test: `runtime/tests/workflow/session-adapters.child-name.test.ts`.
The old derivation was confirmed to fail the same validator before the fix.

### DIFF
--- a/runtime/src/app-server/workflow/session-adapters.ts
+++ b/runtime/src/app-server/workflow/session-adapters.ts
@@ -219,6 +219,21 @@ export {
 } from "./child-terminals.js";
 
 /**
+ * The agent name a workflow child is registered under.
+ *
+ * `assertValidAgentName` accepts lowercase letters, digits and underscores
+ * only, while a child run id carries dashes, colons and a `#` (for example
+ * `wf-3f78249a-...:plan#1`). Folding those to underscores is what keeps the
+ * two ends agreeing: without it every spawn was rejected as "agent_name
+ * must use only lowercase letters, digits, and underscores", which failed
+ * `workflow.plan` on both attempts and ended the run with
+ * `step_retries_exhausted` before any work was done.
+ */
+export function workflowChildAgentName(childRunId: string): string {
+  return childRunId.toLowerCase().replace(/[^a-z0-9_]+/g, "_");
+}
+
+/**
  * The child usage rollup for a real delegate spawn, resolved from the
  * durable source that already exists: the admission kernel reconciles
  * ACTUAL provider usage per reservation for the child run's admissions.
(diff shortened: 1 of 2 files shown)


signerless:
Both red jobs are pre-existing on `main`, not effects of this change.

**default-suite 3/4** — `tests/fnd/benchmark-baselines.contract.test.ts`:

```
Error: current production module closure for csv_scheduler_progress_scan is stale
```

That check compares the recorded module digests against the files on disk. This branch touches one file, `runtime/src/app-server/workflow/session-adapters.ts`, which is in neither closure — 37 modules for `csv_scheduler_progress_scan`, 5 for `patch_delete_parser_historical_comparison`, none of them mine.

Reproduced on a clean checkout of `origin/main` (236673825), with nothing from this branch applied:

```
npx vitest run tests/fnd/benchmark-baselines.contract.test.ts
Tests  2 failed | 7 passed (9)
"Error: current production module closure for csv_scheduler_progress_scan is stale"
```

The baseline needs regenerating on main; I have left it alone rather than record one from an unrelated machine.

**macos-native** — `tests/fnd/red-probe-runner.contract.test.ts` timed out at 15000ms on a fixture unrelated to workflow spawning.

The workflow suite this change belongs to is green: `tests/workflow`, 13 files, 95 tests, all passing, including the new one.

PR #1772, opened 2026-08-24
Fix local models: stop discarding their answers, and stop lying about the window

signerless:
Five bugs that stacked into "local models don't work". In every case the
model was fine; the daemon was not, and the app blamed the model.

### 1. The turn's answer was discarded by budget bookkeeping

The admission step id is built from `turnCount` and `recoveryReentryCount`.
Neither advances on most re-entries — `turnCount` moves only in commit,
which the transition path skips, and `recoveryReentryCount` moves only when
the fallback ladder reserves. A mid-turn compaction, an empty-response
retry, or a second transition after a reserved one re-entered sampling on
an identity that was already spent. `acquire` compares the persisted
request against the new one, the message list has grown by then, and it
rejects the call with `AdmissionStepConflictError`.

By that point the model had streamed its full answer. The turn ended with
`lastAgentMessage: ""`:

```
adm  queued/allowed/dispatched/reconciled  model:sub-…-2:1:0:primary
ANSWER: I'm ready — just say the word and I'll do whatever you need.
!! WARNING stream_model_failed: AdmissionStepConflictError: admission step
   identity already exists with different request data: …:1:0:primary
turn_complete last= ""
```

Sampling rounds are now counted at the one point every re-entry funnels
through, and the step id is keyed on that. The counter rides the checkpoint
slice with the others, so resume reproduces the identity it crashed on.
Two rounds of the same turn now read `…:1:0:primary` and `…:1:1:primary`,
and both are admitted.

### 2. The context window was off by 16x

LM Studio's OpenAI-shaped `/v1/models` returns ids and nothing else, so a
local model fell through to models.dev and inherited whatever window the
upstream weights advertise. A `qwen3.8-27b` loaded at **8192** was treated
as a **128k** model, so nothing trimmed and nothing warned.

Its native listing carries the number: `loaded_context_length` when the
model is loaded, `max_context_length` otherwise. Other local servers keep
the OpenAI-shaped path they already used.

### 3. …and then the output reservation filled what was left

Output is reserved out of the same window as the prompt, and admission
refuses the call when the two do not both fit. The 32k default assumes a
model with room to spare; against a 32k local window there was nothing left
and every turn died on `context_window_exceeded` before anything reached
the model — including the compaction that might have made room.

The default is now a quarter of the known window. A quarter rather than a
half because an agent turn is mostly input: a measured first turn against
qwen3.8-27b counted **17,658 tokens** of system prompt and tool schemas
alone. Only the default is bounded — escalation is a deliberate act against
a turn already known to need the room and keeps its own ceiling, and an
explicit `max_output_tokens` stays authoritative. Cloud windows are far
wider than the 32k default, so this only binds where it must.

### 4. The denial would not say by how much

`context_window_exceeded` on its own sends people looking at their prompt.
It now reports both sides and the window:

```
context_window_exceeded (input 17658 + reserved output 16384 > window 31129)
```

That one line is what located #3.

### 5. The reason never left the daemon

`error` and `stream_error` were on the canonical live-event allowlist;
`warning` was not. The explanation for an answerless turn reached the
rollout and no further, so every live client had to guess — agenc-desktop
guessed "local llama.cpp servers reject its tool schemas", which was almost
never true. The TUI already renders warnings on replay; it just never got
one live.

### Verified

Against a real LM Studio serving qwen3.8-27b through AgenC Desktop. Before,
the answer was journalled and dropped and the composer showed nothing.
After, admission passes, both sampling rounds of a turn are admitted under
distinct identities, and when a turn genuinely cannot run the banner quotes
the daemon's own words.

Six tests fail on `main` too and are untouched here: the Groq/Mixtral route
listing, three `createProvider` env-leak cases, the fable-5 thinking config,
and the Gemini function-declaration case.

### Not fixed here

Core's own token estimate runs well above what the server reports — 26,852
estimated against 7,277 counted by LM Studio on the same turn. That is what
still denies the mid-turn compaction step on small windows, and it deserves
its own change.


signerless:
Update 2026-08-28: pushed 8 additional reliability commits through 578d7dad88335b362fc0f297780643a57f18028b. This includes deterministic pre-write failure settlement, authoritative effect receipts, recoverable nonterminal tool errors, run epoch handling, deterministic tool refusals, OpenAI stream watchdog recovery, unified exec cwd/EPIPE containment, and durable effect-accounting recovery.\n\nLocal validation on Node 26: 163 focused admission/restart/schema tests passed; 4 unified-exec crash regressions passed; 2 SIGKILL tool-ack recovery cases passed; runtime typecheck passed; Core build passed and dist/VERSION points at 578d7da. Independent audit completed with no remaining HIGH/MEDIUM findings.\n\nA clean Electron E2E with OpenAI gpt-5.6-sol is currently running in /private/tmp/agenc-long-e2e.4PAkBx. At handoff it had passed 20 minutes, 107/107 settled tool calls, no disconnect, no crash and no resolve marker. Core PR must remain unmerged until the receiving agent completes or restarts the full 50-minute run.

PR #1775, opened 2026-08-27
refactor(runtime): establish canonical runtime authority

7etsuo:
## Summary

- Removes parallel configuration, hook, provider/model, credential, MCP, workflow, and runtime-state authority paths.
- Makes provider connections consume prepared bindings while preserving provider identity separately from OpenAI-compatible transport.
- Makes bare mode suppress every hook family, repairs OAuth refresh, and binds consent and trust decisions to canonical persisted state.
- Fixes repeated-launch resume, question-mark and Escape input handling, status/config reporting, doctor probes, and durable direct-shell state.
- Centralizes shell, temp, plugin, bundled-skill, prompt, tool-discovery, and untrusted-hook authority.
- Deletes retired provider-selector environment names and obsolete documentation, and aligns theme and native secure-storage naming.
- Binds macOS Keychain reads and writes to one captured user-domain search list, with fail-closed target selection.
- Isolates external providers (grok, openai-compatible, and every non-Anthropic slug) from Anthropic credentials. They no longer read `ANTHROPIC_*` environment variables or native secure storage, provider identity is decided before any credential-storage access (including ahead of the `agenc ssh` auth-proxy branch), and Anthropic-only startup notices are skipped. This removes the synchronous Windows DPAPI PowerShell spawn from the TUI's first render, which is where the hosted `neovim-win-x64` lane stalled blank.
- Keeps the red-probe bootstrap from masking a dependency-provided nonzero exit status (macOS red-probe lane).
- Makes the Linux subreaper broker deliver a graceful stop exactly once. It signalled the root's process group and then the root itself, so a leader scheduled between the two calls saw SIGTERM twice; the hosted default-suite runner (no writable cgroup, so always the broker path) hit this in `supervisedProcess.test.ts`. The broker now skips the direct kill for any process the group delivery already reached; SIGKILL keeps the redundant delivery. A regression test runs 40 subreaper-contained leaders in parallel and requires one delivery each.
- Repins the FND benchmark baseline to the fixed broker source and the `runtime/src` tree at `0997edc39` (the baseline contract binds the broker digest). Note for after the squash-merge: the provenance check requires the pinned revision to be an ancestor of HEAD, so `main` will need a main-tip rebind commit exactly as #1764 did after the previous repin; the branch already carried a topic-branch pin before this work.
- Removes the temporary hosted TUI startup tracing; the affected files are byte-identical to their pre-trace state. The bounded PTY failure output and the nonblank-frame prompt-readiness requirement in the E2E harness stay.

## Verification

Local verification passed at exact SHA `fc93c475e69030322c25b7bb2436d0a34b4c6452`.

- Full Vitest suite under the Docker hermetic boundary with `--require-zero-skips`: 2,140 files and 20,734 tests passed; required-gate tests 133/133, agent-surface contract tests 22/22, launcher tests 191/191, red probes exact (files=1 expected-red=1 skipped=0).
- Bun suite: 60 isolated files passed.
- Runtime typecheck, build, package entrypoints, generated SDK types, and real PTY startup smoke passed (`validate:runtime`, and again in every pre-commit hook run).
- Agent-surface contract passed.
- SPDX SBOM generated and validated: 681 packages and 1,075 relationships.
- SDK build and typecheck passed.
- Daemon, LLM, and real TUI end-to-end suites passed: 14/14, 7/7, and 115/115.
- Hosted-platform PTY scenarios (`check-tui-e2e/runner.mjs --platform linux-x64`): 2/2 passed.
- Locally runnable platform lanes with the exact hosted configs and zero skips: PowerShell 40/40, Neovim platform 65/65 and Neovim lifecycle 18/18 (checksum-pinned Neovim 0.12.1), cross-platform native/FND subset 89/89, red-probe audit exact. The kernel-sandbox lane passes 8/9 locally; the remaining assertion needs the root-loaded `agenc-native-userns` AppArmor profile that only the hosted lane provisions.
- Independent reviews: the delta since the last reviewed SHA (`3dae89946`) was reviewed in five passes (auth and keychain delta, ssh-proxy guard, proxy account-lookup test, broker signal fix, helper guard plus baseline repin). The first pass found no blocker; its one actionable finding (provider identity decided after the ssh auth-proxy branch) is fixed in `349e87234` with a revert-sensitive test, strengthened in `7a0c0b45b`. The broker fix pass confirmed the doubled delivery independently (7 of 120 leaders with the old broker, 0 of 320 with the fix) and reported no blocker. The final pass reports no code defect; its one finding is the post-merge baseline rebind noted above.
- GitNexus refreshed the full index with PDG data and helper skills at this SHA, then compared the branch with `main`: 1,931 files, 9,804 symbols, and 236 execution flows changed. GitNexus reports CRITICAL aggregate scope because of the branch size; it did not report a specific new defect. Per-commit staged scope for every commit added in this pass was low risk with zero affected flows.

Hosted platform matrix: the first exact-head run at `7a0c0b45b` passed 12 of 13 lanes (including `neovim-win-x64`, the first green for that lane on this branch); the one failure was the broker double-delivery above, fixed locally in `07bb314b1`. The final exact-head run at `fc93c475e` is recorded on the PR checks.

https://claude.ai/code/session_01C8yhaVpinooyLjiAjRUpGr


7etsuo:
Closing temporarily for an interactive local-build acceptance test before review.

PR #2038, opened 2026-09-02
test(runtime): retry lost SIGTERM in daemon-cli teardowns

FalconOrtiz:
## Why

Foreground-daemon cases in `daemon-cli.contract.test.ts` emit one `SIGTERM` then `await running`. `runAgenCDaemonCli` installs shutdown handlers only after services are constructed. The test signal process has no default action, so a signal in that window is dropped, the run never settles, and vitest reports a content-free 30s timeout. #2018 fixed one call site on `harness-tools`; the other teardowns share the same hang.

## Scope

- Add `stopRunningDaemons` (and `DAEMON_STOP_GRACE_MS`) in `runtime/tests/app-server/daemon-cli.contract.test.ts`.
- Route every foreground-daemon teardown through it (retry SIGTERM until the run settles, or fail with a named error).
- Leave the intentional mid-test SIGTERM in the lifecycle-lock authority case alone.

Out of scope: the `waitForPid` milestone-budget rewrite from #2018.

## Tradeoffs

Retry lives in the test harness, not production. Production signal delivery still has a default action. The harness double does not, so the retry belongs next to that double.

## Blast Radius

Contract tests only. Reviewers and CI see named stop failures instead of silent vitest timeouts when a teardown races handler install. Main stays flaky on slow runners until this lands.

## Verification

- Synthesized the lost-SIGTERM race against the helper: a single emit before the listener hung; `stopRunningDaemons` settled after 5 SIGTERMs (exit 0).
- Full `daemon-cli.contract.test.ts` on this Windows host is not a valid surface (path/ACL/Unix-socket failures dominate). CI Linux `pr-fast` is the package gate for this file.

### DIFF
--- a/runtime/tests/app-server/daemon-cli.contract.test.ts
+++ b/runtime/tests/app-server/daemon-cli.contract.test.ts
@@ -560,6 +560,66 @@ async function waitForPid(pidPath: string): Promise<number> {
   throw new Error("timed out waiting for daemon pid");
 }
 
+/**
+ * How long {@link stopRunningDaemons} waits for the runs to settle once its
+ * SIGTERM loop has given up. Bounded so a daemon that never stops fails with a
+ * named error instead of the case dying on the vitest deadline.
+ */
+const DAEMON_STOP_GRACE_MS = 2_000;
+
+/**
+ * Shuts foreground daemons down without depending on when they installed
+ * their signal handlers.
+ *
+ * runAgenCDaemonCli only calls installAgenCShutdownSignalHandlers after its
+ * services are constructed. The test signal process has no default action, so
+ * a single SIGTERM emitted before that point is dropped, the daemon keeps
+ * running, and the awaited run promise never settles -- the test then dies on
+ * the vitest deadline with no indication of what it was waiting for. Repeating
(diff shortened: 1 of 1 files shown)


FalconOrtiz:
Pushed `28ec7b8b7` for Sonar **Duplication on New Code** (was 13.7%, gate ≤3% / ship bar **0.0%**).

Collapsed repeated teardown sites into `stopDaemon` / `stopDaemonQuiet` and shortened the shared helper. Multi-daemon teardowns still use `stopRunningDaemons`.

FalconOrtiz:
Pushed `609c2e5e4` to clear Sonar **Duplication on New Code** (was ~6.5%).

Root cause was not missing teardown retries: `createSignalProcess` dropped `SIGTERM` when no listener was installed yet. The PR now queues that signal and flushes on `once()`, with two focused cases. Removed the `stopRunningDaemons` call-site churn that was duplicating identical `finally` blocks across New Code (diff vs main is ~+38/-2).

7etsuo (approved):
Reviewed on current main. The test-only signal double now preserves a startup-raced signal until listener registration, keeps live delivery behavior, and the daemon CLI contract suite passes (124/124).

PR #2039, opened 2026-09-02
fix(runtime): defer SessionStore diagnostics until flush exits

### LINKED ISSUE #2032: a diagnostic emitted during a rollout flush is appended to the batch being flushed, corrupting the journal sequence
## Summary

`SessionStore.flushBatchUntimed` calls `this.emitDiagnostic(...)` on two paths
while it is mid-flush, before `pending` has been drained. When a session has
mounted its rollout store, that diagnostic channel is wired straight to
`session.emit`, which allocates a live EventLog sequence and appends the
warning to the very batch being written. The result is a canonical journal with
two adjacent items carrying the same sequence, which fails resume.

## Where

`runtime/src/session/session-store.ts`:

- line ~2042, `cause: "event_log_batch_delayed"` (the I-83 suspend/resume gap)
- line ~2092, `cause: "rollout_degraded"`

Both sit inside `flushBatchUntimed`, ahead of
`const toWrite = this.pending; this.pending = []`.

The wiring that makes it reentrant is `Session.mountRolloutStore`
(`runtime/src/session/session.ts`, around line 5231), which points the store's
diagnostic channel at `session.emit`.

## How this was found

A third caller of the same channel was added on the `harness-daemon` branch:
`flushBatch` was wrapped in the slow-store-op guard with a custom reporter, so
any flush slower than `SLOW_STORE_OP_THRESHOLD_MS` (50 ms,
`runtime/src/utils/slow-store-op.ts`) emitted through it. On the ubuntu-24.04
runner a durable fsync crosses 50 ms; on an M-series Mac it does not. CI failed
with:

```
CanonicalJournalIntegrityError: canonical journal repeats its preceding event sequence
  at StrictCanonicalJournalValidator.#validateSequence  src/state/recovery-journal-contract.ts:1170
  at scanPass                                          src/session/canonical-rollout-scanner.ts:1133
```

Reproduced locally by setting `SLOW_STORE_OP_THRESHOLD_MS = -1` so every flush
trips it. A journal dump then shows the injected warning taking sequence 2
immediately before the next event that also carries sequence 2, and the write
is itself flushed durably, so a disk slow enough to trip the threshold keeps
tripping it: warnings cascaded at sequences 2, 3 and 4.

That third caller is removed (the flush timing now goes to the process-wide
sink). **These two paths are untouched and predate that branch.**

## Why it still matters

Neither remaining path fires on disk speed, which is what made the third one so
easy to hit, but both are reachable in production:

- `event_log_batch_delayed` fires when the process was suspended and resumed
  mid-batch, which is ordinary laptop-sleep behaviour.
- `rollout_degraded` fires when the durable write failed and items were
  requeued, i.e. precisely when the journal is already under stress.

In both cases the diagnostic is appended into the batch currently being
written, so the failure mode is a corrupted journal that only shows up later,
at resume, as an integrity error rather than at the moment of the fault.

## Suggested fix

Do not emit into a live-sequenced channel from inside the flush. Either:

1. Queue the diagnostic and emit it after `pending` has been drained and the
   write has completed, or
2. Give `SessionStore` a diagnostic channel that cannot allocate an EventLog
   sequence, and let `mountRolloutStore` bridge to `session.emit` only for
   diagnostics raised outside a flush.

Whichever shape, the invariant worth pinning in a test is: a diagnostic raised
during a flush never lands in the batch that flush is writing.

## Tests

- A flush that trips `event_log_batch_delayed` leaves the journal sequence
  contiguous and passes `scanCanonicalRollout`.
- The same for `rollout_degraded` with a failing durable write.
- Both should fail before the fix; the reproduction recipe above (forcing the
  threshold) shows how to make the timing deterministic.

## Related

`SLOW_STORE_OP_THRESHOLD_MS` is a hard-coded 50 ms with no env override, which
is why reproducing this needed a source edit. Making it configurable would make
this whole class of timing-dependent CI red reproducible without patching src.

FalconOrtiz:
## Why

A diagnostic raised inside `SessionStore.flushBatch` can re-enter through `Session.mountRolloutStore`, which wires the channel to `session.emit`. That appends a live-sequenced event into the batch currently being written and corrupts the canonical journal sequence (#2032).

## Scope

- `runtime/src/session/session-store.ts`: defer `emitDiagnostic` while `flushDepth > 0`, then deliver after the flush unwinds. Emit `rollout_degraded` only on the degraded-mode transition.
- `runtime/tests/session/session-store-flush-diagnostic-reentry.test.ts`: regression for the I-83 and `rollout_degraded` paths.

Out of scope: making `SLOW_STORE_OP_THRESHOLD_MS` configurable.

## Tradeoffs

Deferred diagnostics still reach the listener after the flush. They no longer share the in-flight batch. `rollout_degraded` fires once per enter instead of on every failed write while already degraded.

## Blast Radius

Session stores with a diagnostic listener that appends (production `mountRolloutStore`). Resume integrity for I-83 suspend markers and disk-degraded flushes.

## Verification

- `node scripts/run-hermetic-vitest.mjs run tests/session/session-store-flush-diagnostic-reentry.test.ts` exit 0 (2 pass; failed before the fix with the diagnostic id present in the first flush)
- `node scripts/run-hermetic-vitest.mjs run tests/session/session-store-flush-diagnostic-reentry.test.ts tests/session/session-store-i83-suspend-drop.test.ts` exit 0 (3 pass)

Fixes https://github.com/tetsuo-ai/agenc-core/issues/2032

### DIFF
--- a/runtime/src/session/session-store.ts
+++ b/runtime/src/session/session-store.ts
@@ -1421,6 +1421,14 @@ export class SessionStore {
   private readonly diagnosticsBuffer: SessionStoreDiagnostic[] = [];
   private onDiagnostic?: (d: SessionStoreDiagnostic) => void;
   /**
+   * Depth of an in-flight `flushBatch`. Diagnostics raised while > 0 are
+   * deferred so a listener wired to `session.emit` (see
+   * `Session.mountRolloutStore`) cannot append a live-sequenced event into
+   * the batch currently being written (#2032).
+   */
+  private flushDepth = 0;
+  private readonly deferredFlushDiagnostics: SessionStoreDiagnostic[] = [];
+  /**
    * I-38 async fsync retries currently in flight. Tracked so `close()`
    * can wait for them to settle (or so tests can await completion).
    * Each entry is the Promise returned by `scheduleFsyncRetry()`.
@@ -1812,12 +1820,26 @@ export class SessionStore {
     for (const d of buffered) listener(d);
   }
 
-  private emitDiagnostic(d: SessionStoreDiagnostic): void {
+  private deliverDiagnostic(d: SessionStoreDiagnostic): void {
     if (this.onDiagnostic) {
       this.onDiagnostic(d);
     } else {
       this.diagnosticsBuffer.push(d);
     }
+  }
+
+  private emitDiagnostic(d: SessionStoreDiagnostic): void {
+    if (this.flushDepth > 0) {
+      this.deferredFlushDiagnostics.push(d);
+      return;
+    }
+    this.deliverDiagnostic(d);
+  }
+
+  private releaseDeferredFlushDiagnostics(): void {
+    if (this.flushDepth !== 0) return;
+    const deferred = this.deferredFlushDiagnostics.splice(0);
+    for (const d of deferred) this.deliverDiagnostic(d);
   }
 
   /** Drain any buffered diagnostics. Used by tests. */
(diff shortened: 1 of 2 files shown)


FalconOrtiz:
Pushed `1885c0ff5` for Sonar duplication (was 6.7%).

`flushBatch` is now a thin flush-depth wrapper around `flushBatchUnlocked` (avoids re-indenting the whole body as new duplicated code) and the regression tests share one live-diagnostic helper.

7etsuo (approved):
Reviewed on current main. The semantic change is limited to deferring diagnostics through the outermost flush boundary and emitting the degraded diagnostic once per transition. All session-store tests pass (67/67), both TypeScript configurations pass, and hosted affected tests are green.

Issue #2023, opened 2026-09-02
State that should survive a daemon restart does not: extraction cadence resets, extraction child cannot reach the global memory root, resumed session keeps its first compaction limit


signerless:
## Summary

Three independent runtime defects seen in the 15-prompt AgenC Desktop run (session `conv-mtjdmlfc`, Grok 4.6, 2026-09-02) that survive a daemon restart badly. Each part below has its own root cause, files, fix and tests; they can be fixed in separate PRs.

## Part A. Memory extraction cadence restarts with the daemon

**Symptom.** Extraction runs every third eligible turn (`minEligibleTurns` cadence). The counter lives in process memory, so after every daemon restart it starts again at 1/3. The run had three restarts (all for unrelated resume fixes) and the extraction ran once in 13 turns: prompt 8 logged `memory_extraction` running (3/3), prompt 11 after a restart logged `memory_extraction_skipped: deferred by eligible-turn cadence (1/3 eligible turns)`, and so on.

**Where.** `runtime/src/services/extractMemories/extractMemories.ts`: `initExtractMemories` keeps `ExtractionLane` objects (`extractionLane(session, memoryDir)`, `lane.trigger.processedVisibleCount`, eligible-turn counting, `pruneIdleLanes`) in a module-level map.

**Fix.** Persist the cadence state per session and restore it on resume: store `processedVisibleCount` and the eligible-turn count in the session state (the `session_state` rollout item or the SQLite session snapshot), and seed the lane from it when the session is reconstructed. Alternative with the same effect: decide by history, not by process life (run extraction when the number of model-visible messages since the last successful extraction reaches the threshold; the last extraction point can be stored as a marker in the rollout).

**Tests.** Two `initExtractMemories` instances over the same restored session state: the second continues the count instead of restarting; a restart between the second and third eligible turns still yields an extraction on the third.

## Part B. The extraction child cannot use the global memory root

**Symptom.** The child probed the global memory directory and was refused:

```
FileRead /Users/.../agenc-home/memory/MEMORY.md
-> {"error":"FileRead is restricted to the memory directory: /Users/.../agenc-home/projects/<slug>/memory/"}
```

The extraction prompt's taxonomy includes `user` type memories (who the user is), which belong to the global root, but the child can only read and write the project root, so such memories can never be extracted there. The main agent's memory prompt lists both roots; the child sees one.

[...]

PR #2040, opened 2026-09-02
fix(runtime): persist memory extraction cadence across restart

### LINKED ISSUE #2031: memory extraction cadence resets on daemon restart because it lives only in process memory
## Summary

The memory-extraction eligible-turn cadence is counted in process memory. A
daemon restart clears the counter while the conversation it paces stays on
disk, so the session waits a further full cadence before any memory is
written. This is Part A of #2023, deliberately left undone by #2026.

## Where

`runtime/src/services/extractMemories/extractMemories.ts` keeps an
`ExtractionLane` per `(conversationId, memoryRoot)` in a module-level map.
`lane.trigger` is a `MemoryExtractionTriggerState`
(`runtime/src/memory/extraction-triggers.ts`) holding `processedVisibleCount`
and `turnsSinceLastExtraction`, both created at zero by
`createMemoryExtractionTriggerState()` in every new process.

`shouldDeferForEligibleTurnCadence` decides purely from
`turnsSinceLastExtraction` against `minEligibleTurns` (default 3).

## Impact

Bounded but real: after a restart, up to `minEligibleTurns` eligible turns pass
before extraction runs again. Nothing is lost, because `processedVisibleCount`
resetting to zero means the next run sees the whole conversation as
unprocessed. It is a delay, not a gap.

## What was already tried, and why it is not in the tree

#2026 first added a bypass: the first cadence decision a process made for a
lane could consult the unprocessed backlog instead of the counter, on the
theory that a backlog much larger than one turn's output means the process
inherited the conversation. That is wrong. A single turn with tool calls and
attachments produces well over the eight-message threshold it used, so the
bypass fired on the first turn of every new session and launched a
full-history extraction child there.

**Do not reintroduce a message-count heuristic.** Counts cannot distinguish
"this process inherited a conversation" from "this process just built one".

## Suggested fix

Persist the cadence state per session and seed the lane from it when the
session is reconstructed. Both fields matter: `processedVisibleCount` (so a
restart does not re-extract the whole conversation) and
`turnsSinceLastExtraction` (so the pacing continues).

### Trap to avoid

The obvious slot is a `session_state` rollout item. Note
`latestPersistedAgentTask` in
`runtime/src/session/agent-task-lifecycle.ts`:

```ts
for (let i = rolloutItems.length - 1; i >= 0; i -= 1) {
  const item = rolloutItems[i];
  if (item && item.type === "session_state") {
    return { value: item.payload.agentTask };
  }
}
```

It returns on the first `session_state` item it meets walking backwards and
reads that item's `agentTask`. `SessionStateUpdate` currently has exactly one
slot. Writing a `session_state` item that carries only a memory-extraction slot
therefore reads as "the agent task was explicitly cleared", and the persisted
agent task is dropped on the next restore. `undefined` does not survive
`JSON.stringify`, so key presence cannot be used to tell the cases apart after
a round trip.

Either write both slots on every update (read-modify-write), or make each
walker skip items that do not carry its own slot, and cover that with a test
that persists one slot and asserts the other survives.

## Tests

From #2023 Part A, unchanged:

- Two `initExtractMemories` instances over the same restored session state: the
  second continues the count instead of restarting it.
- A restart between the second and third eligible turns still yields an
  extraction on the third.

Plus, for the trap above: persisting extraction cadence does not clear a
persisted agent task.

## Not in scope

The cadence default itself (3) and the lane keying are fine. This is only about
surviving a restart.

FalconOrtiz:
## Why

Memory extraction cadence lived only in the in-process lane map. A daemon restart zeroed `turnsSinceLastExtraction` and `processedVisibleCount` while the conversation stayed on disk, so the session waited another full cadence before writing memory. This is #2031 (Part A of #2023 left after #2026).

## Scope

- Add `memoryExtractionTrigger` on `SessionStateUpdate` and persist it from `extractMemories` after each cadence or cursor change.
- Seed a new `ExtractionLane` from restored session state on resume via `recordInitialHistoryOnResume`.
- Make `session_state` walkers slot-aware (`latestPersistedSessionStateSlot`) so a memory-only update cannot clear `agentTask`. Clear agent tasks with explicit `null` so the key survives JSON.
- Schema, reducer, and regression tests for restart cadence and the agent-task trap.

Out of scope: changing the default cadence of 3, lane keying, or message-count bypass heuristics.

## Tradeoffs

Slot-aware walkers instead of read-modify-write on every writer. Writers stay independent; clear must use `null` rather than omitted/`undefined`.

## Blast Radius

Resume and memory extraction paths only. Existing rollouts without the new slot keep today's zero seed. Agent-task clear now writes `null`; restore already treated missing values as clear.

## Verification

- Focused cadence and slot-trap tests in `extractMemories.test.ts` and `agent-task-lifecycle.test.ts`: exit 0
- Full `agent-task-lifecycle` + `recovery-journal-contract` suites: 45 passed, exit 0
- Node 26.7.0. Pure unit tests on Windows host (no daemon suite).

Fixes #2031


FalconOrtiz:
**Merge request:** conflicts with `main` resolved (`3ec5ebc8a`). Please squash-merge when Sonar/CI green. FalconOrtiz cannot merge.

FalconOrtiz:
QG passed with **1 New issue** (S6582 optional chain in `rollout-item.ts`). Cleared for excellency bar (0 OPEN). Pushed for re-analysis.

7etsuo (changes requested):
This branch is not type-clean and has two cadence-state correctness issues.

1. `runtime/src/state/recovery-journal-schema.ts:1449` fails Node 26.5 TypeScript with TS2322 because the `session_state` object validator infers `Validator<unknown>` slots and no longer satisfies the exact `SessionStateUpdate` contract. This reproduces locally and in pr-fast.

2. `extractionLane` became async and checks `lanes.get(key)` only before `await readSeededTrigger(session)`. Two simultaneous first requests can both observe no lane, yield, then create separate `ExtractionLane` objects. That bypasses the existing `inProgress`/`pendingContext` coalescing contract. Re-check or memoize initialization after the await and add a simultaneous-first-request regression.

3. The live lane key is session plus `memoryRoot(memoryDir)`, but the persisted `memoryExtractionTrigger` contains no root identity. After a memory-directory/config change and restart, the new root can inherit another root’s `processedVisibleCount` and skip history that was never extracted there. Persist per-root state or bind and validate the root before restoring.

Please also validate the counters as non-negative safe integers at the recovery boundary rather than arbitrary finite numbers.

FalconOrtiz:
Addressed CR blockers on `c13d073bf`:

1. Exact nullable `session_state` validator typing (clears TS2322 Validator vs never).
2. ExtractionLane recheck after await — concurrent init shares the created lane.
3. Persisted cadence scoped to normalized memory-root identity; foreign roots do not seed.

Local e2e smoke: `e2e_smoke_ok` — `e2e-smoke-2040-c13d073b.md` (runtime typecheck + extractMemories cadence regressions). Ready for re-review / squash-merge.

FalconOrtiz:
Cleared the Sonar **1 New issue** (`typescript:S6582`): prefer optional chain on the cadence memory-root guard.

`if (stored?.memoryRoot !== memoryRoot(memoryDir))` — same semantics, zero open new issues target.

Head: `65a129c83655f0c44dcec54230ba3f10f1793ecc`. Local e2e smoke re-run after this push.

FalconOrtiz:
**Requesting squash-merge** when you are satisfied with re-review.

Blockers from the prior CHANGES_REQUESTED pass are addressed on `65a129c83`. Local `e2e_smoke_ok` for that head. Validator/race/memory-root + Sonar S6582; e2e_smoke_ok

I do not have merge rights on `tetsuo-ai/agenc-core` — please squash-merge if this looks good.

7etsuo (changes requested):
Request changes at `65a129c83655f0c44dcec54230ba3f10f1793ecc`.

This branch conflicts with current main in six files, including extraction, agent-task lifecycle, rollout items, and the recovery schema. More importantly, #2143 already landed durable extraction cadence through `runtime/src/session/memory-extraction-state.ts` and the `memoryExtraction` slot. It keeps progress per normalized memory root and restores the latest value for each root. This PR introduces a competing `memoryExtractionTrigger` slot with a single current-root value. Replaying A, then B, and restarting on A loses A's saved progress in this implementation.

Rebase onto the existing main implementation and identify any remaining behavior this PR needs to fix. Preserve the root-keyed state and the later bounded-batch/retry behavior from #2164. If no gap remains, this can be closed as superseded instead of adding another persisted representation.

The new recovery validator also still accepts arbitrary finite numbers for both counters (`recovery-journal-schema.ts:254-255`). Main's existing implementation requires nonnegative safe integers; preserve that validation when rebasing.

The earlier post-await lane recheck and slot-aware agent-task restoration are present. Runtime typecheck and the extraction, lifecycle, and recovery-schema suites pass on this PR's own head; that does not validate a resolved merge with current main.

FalconOrtiz:
Closing as superseded by #2143, which is now on `main`.

#2143 includes the durable extraction cadence, slot-aware agent-task/reducer handling, restart restoration, and regression coverage from this PR, while using root-keyed `memoryExtraction` state so progress survives switching between memory roots. It also validates both counters as nonnegative safe integers and restores malformed persisted data defensively.

The remaining differences here do not add independent value: the lane double-check compensates for this branch's asynchronous lane creation (main creates the lane synchronously and restores under its in-progress guard), and the nullable validator helpers belong to the superseded single-trigger payload API.

Thanks for the work that established and tested the restart-cadence fix.

PR #2041, opened 2026-09-02
fix(runtime): keep thread_rollout_items paced with live flushes

### LINKED ISSUE #2028: thread_rollout_items drifts behind the rollout, and threads.last_item_index is vestigial
## Summary

The thread store holds only a prefix of a session's rollout, and the importer's cursor never advances. Any read that goes through the persisted path — which is what happens whenever a session's agent is not live — returns a projection of that stale prefix, with recent messages silently missing and no error.

Found while verifying the transcript truncation in tetsuo-ai/agenc-desktop#156. That issue has a separate, sufficient cause (compaction was treated as a transcript boundary, fixed in #2027); this is a second, independent route to an incomplete transcript and needs its own fix.

## Evidence

Measured on a live project database, `agenc-state_1.sqlite`, for a conversation whose rollout on disk is 14,223 lines:

```sql
select thread_id, min(item_index), max(item_index), count(*)
  from thread_rollout_items where thread_id='conv-mtjdmlfc';
-- conv-mtjdmlfc | 0 | 3969 | 3970
```

So 3,970 of 14,223 items were imported, and the import stopped at index 3969.

```sql
select thread_id, last_item_index from threads;
-- every one of the 28 rows reads 0
```

`threads.last_item_index` is 0 for every thread in the database, including threads whose items were imported. That cursor is what an incremental import would advance, so either nothing advances it or the advance is not persisted.

## Where to look

- The thread-store importer and whatever owns `threads.last_item_index`: find the writer, and whether it runs once at import or incrementally as the rollout grows.
- `#readPersistedSessionTranscriptV2` in `runtime/src/app-server/background-agent-runner.ts`, the fallback used when a session's agent is not live. It projects whatever the store holds, so a short store silently yields a short transcript.
- Whether the importer is meant to keep pace with a growing rollout at all, or whether the persisted path should read the rollout file directly when the store is behind.

## Questions to answer

1. Is `last_item_index` dead, or written somewhere that never commits? A cursor reading 0 for 28 rows including fully imported ones points at the former.
2. What stops the import at its observed end? 3,970 is not a round number; check for a batch bound, a transaction that ends on first error, or an import that only runs once at thread creation.
3. Should the persisted transcript path detect that the store is behind the rollout and fall back to the file, rather than returning a short answer as if it were complete? Silently returning a prefix is the part that makes this hard to notice.

## Acceptance

- For a session with a long rollout and no live agent, the persisted transcript path returns the same messages as the live path.
- `threads.last_item_index` reflects what was actually imported, and a second import resumes from it rather than redoing or skipping work.
- A regression test with a rollout longer than whatever bound is found, asserting the projection covers it.

FalconOrtiz:
## Why

`thread_rollout_items` lagged the on-disk rollout because live sessions write through `RolloutStore` and never hit `FileThreadStore.appendItems`. Daemon run replay reads that mirror read-only, so a stale projection silently truncates journal pages. `threads.last_item_index` was never read or written; freshness already lives in `backfill_files`.

## Scope

- `SessionStore.setOnRolloutCommitted` fires after a successful flush write.
- `FileThreadStore` binds that hook on create/resume, indexes the current file, and clears the hook on shutdown/discard/close.
- Migration `031_drop_threads_last_item_index` drops the vestigial column.
- Regression tests cover live-recorder appends without `appendItems` and assert the column is gone.

Out of scope: transcript `loadHistory` already reads the rollout file (see #2028 comment); no change there.

## Tradeoffs

Writer-side indexing keeps the daemon's read-only replay path correct. A read-time refresh in run inspection cannot mutate state DBs opened read-only.

## Blast Radius

Touches every live `FileThreadStore` session flush. Incremental backfill makes unchanged files a size/mtime no-op. Migration is additive-drop only.

## Verification

- `node scripts/run-hermetic-vitest.mjs run tests/session/thread-store.test.ts -t "indexes appends|last_item_index"` exit 0 (2 pass)
- `node scripts/run-hermetic-vitest.mjs run tests/state/migrations.test.ts -t "loads state migrations|keeps versioned"` exit 0
- `node scripts/run-hermetic-vitest.mjs run tests/state/backfill-incremental.test.ts -t "does not re-INSERT|skips re-indexing"` exit 0

Closes #2028


FalconOrtiz:
**Merge request:** conflicts with `main` resolved (`52848b0cc`). Please squash-merge when Sonar/CI green. FalconOrtiz cannot merge.

7etsuo (changes requested):
There is a durability bug in the new callback placement.

`runtime/src/session/session-store.ts:2137` invokes `onRolloutCommitted` inside the same `try` that classifies append/fsync failures. `FileThreadStore` binds that callback to synchronous file parsing and SQLite writes. If that indexing work throws after `writeBytes*` has already appended the JSONL, `flushBatch` now reports an append failure; an EIO/ENOSPC-style error can enter degraded mode with `safeToRequeue=true`, requeueing rows that are already on disk, and other callback errors escape after the canonical write succeeded. A later retry can therefore duplicate rollout events or surface a false write failure.

Please isolate mirror-index failures from canonical rollout commit accounting (and add a regression where the listener throws after the write). The mirror may lag or report its own diagnostic, but it must not make already-written canonical rows eligible for re-append.

FalconOrtiz:
Addressed CR on `8963dc256`:

- `onRolloutCommitted` is best-effort after canonical JSONL append; callback exceptions no longer fail/requeue the flush (no duplicate JSONL risk).
- Synced migration collision as `032`.
- Regression: callback throw → `flushDurable` succeeds; second flush does not re-invoke.

Local e2e smoke: `e2e_smoke_ok` — `e2e-smoke-2041-8963dc25.md`. Ready for re-review / squash-merge.

FalconOrtiz:
**Requesting squash-merge** when you are satisfied with re-review.

Blockers from the prior CHANGES_REQUESTED pass are addressed on `8963dc256`. Local `e2e_smoke_ok` for that head. onRolloutCommitted best-effort; e2e_smoke_ok

I do not have merge rights on `tetsuo-ai/agenc-core` — please squash-merge if this looks good.

7etsuo (approved):
Reviewed `8963dc256a9ebf706d152da81b06c4ccfc8bb940`. No blocking finding remains.

The mirror callback now runs only after the canonical append succeeds, and its exception is contained so a SQLite indexing failure cannot requeue an already committed batch. The current-writer identity check and listener cleanup cover replacement, shutdown, discard, and close. The incremental backfill path reads the appended tail. Migration 032 removes the unused column without changing the canonical rollout format.

The earlier callback-failure review concern is resolved, with a regression proving that a thrown mirror callback does not duplicate the canonical write.

Validation: the three candidate PRs were combined on current main in integration commit `8587f0e045ba11e031ce45bbe49d06abb797713d`. The combined tree passes 285 focused tests, runtime typecheck, build, PTY startup, and all nine native platform jobs. The restored OAuth credential helper also passes its own suite.

The complete hosted default suite has three green shards. Shard 3 has one failed benchmark binding for `runtime/package.json` and four Darwin-only skips under Linux; both are also present on unchanged main, and the relevant files are identical between main and this integration. No new failure remains in the candidate tree. This is not a claim that main's full suite is green.

Evidence: [final candidate validation](https://github.com/tetsuo-ai/agenc-core/actions/runs/34098153762), [unchanged main baseline](https://github.com/tetsuo-ai/agenc-core/actions/runs/34096518321).

PR #2043, opened 2026-09-02
fix(tui): fail closed on BUFFER input while the provider opens

### LINKED ISSUE #1932: fix(tui): capture focused BUFFER input while the provider is opening
## Problem

The hosted `default-suite 3/4` lane for PR #1917 at exact head `5ab3dcb689eeb5cd005ed9dae44ef799babdc051` failed this existing BUFFER input-ordering contract:

```text
buffer-surface.test.tsx > captures focused vim insert text before later input handlers
expected leaked input []
received ["q"]
```

The PR changes only SDK transcript documentation, a generated SDK type mirror, its checker, and its contract test. It does not touch the TUI input path. An isolated rerun passed, so this is a pre-existing timing race exposed by hosted load.

The source has a concrete fail-open window. `BufferSurface` registers an ordinary `useInputCapture`, but its normal path returns `store.handleInput(...)`. `BufferProviderController.handleInput` returns `false` while no provider is installed. The top-level chord interceptor then leaves the input event unstopped, so a later `useInput` handler can receive a key typed while the focused BUFFER surface is still opening or replacing its provider.

The current test waits a fixed 100 ms after render. On the hosted failure, opening had not reached the provider-ready boundary when the first `q` arrived. The focused editor did not mutate, but the later handler received the key. In production that can put editor input into the composer or trigger another lower-priority handler.

## Suggested fix

Make focused BUFFER ownership explicit across provider startup and replacement.

- When BUFFER is focused and its provider is absent, opening, or being replaced, consume ordinary keyboard input at the top-level capture boundary instead of returning `false`.
- Do not queue and replay text into a later provider or a newly selected file. Consuming it is the safer boundary because the original target and editor mode are not established.
- Preserve the existing ordering: modal recovery captures run first; resolved keybindings and emergency/global actions keep their reviewed behavior; wheel input outside the BUFFER node may still fall through.
- Keep ready-provider input delegated to `store.handleInput`, including its Vim mode and paste semantics.
- Express the loading/transition ownership as a typed controller or surface state instead of inferring it from a timeout.

## Regression coverage

Replace the timing-dependent proof with controlled deferred provider startup:

1. Render a focused BUFFER surface with a later raw input handler.
2. Hold provider open on a deferred promise and send printable, Vim-command, paste, and navigation input before it resolves.
3. Prove none reaches the later handler, no buffer is mutated, and no command is replayed after readiness.
4. Resolve startup and prove ready inline/Neovim input still reaches the provider once.
5. Repeat during provider replacement and file navigation, and retain the outside-wheel fallthrough contract.
6. Run the contract under the default-suite sharding/concurrency settings so the test does not depend on a 100 ms sleep.

Do not fix this by increasing sleeps or retrying the assertion; the product boundary should fail closed while the focused editor is not ready.

FalconOrtiz:
## Why

Focused BUFFER could return false from its top-level input capture while the editor provider was still opening or being replaced. Later useInput handlers then received keys typed for the focused editor, which is the hosted race behind #1932.

## Scope

- Add `BufferOrdinaryInputOwnership` and `getOrdinaryInputOwnership()` on `BufferProviderController`.
- Fail closed in `handleInput` and `BufferSurface` capture while ownership is `transition`, including the host-path gap before `open` starts.
- Replace the 100 ms sleep contract with a deferred selection-factory proof; cover open and replacement at the controller boundary.

## Tradeoffs

Keys typed during open are discarded rather than queued. Replay would target an unknown editor mode and file.

## Blast Radius

TUI BUFFER keyboard ownership only. Ready-provider Vim, paste, and outside-wheel fallthrough stay on the prior paths.

## Verification

- `node scripts/run-hermetic-vitest.mjs run tests/tui/workbench/buffer-surface.test.tsx -t "consumes focused BUFFER input"` (exit 0)
- `node scripts/run-hermetic-vitest.mjs run tests/tui/workbench/buffer-provider-boundary.contract.test.ts -t "consumes ordinary input|conservative fallback"` (exit 0)
- `node scripts/run-hermetic-vitest.mjs run tests/tui/workbench/buffer-surface-handlers.test.tsx` (exit 0)

Fixes #1932

### DIFF
--- a/runtime/src/tui/workbench/buffer/providers/BufferProviderController.ts
+++ b/runtime/src/tui/workbench/buffer/providers/BufferProviderController.ts
@@ -51,6 +51,16 @@ import { emptyProviderSnapshot, INLINE_BUFFER_CAPABILITIES } from "./types.js";
 
 type SelectionFactory = () => Promise<BufferProviderSelection>;
 
+/**
+ * Who owns ordinary keyboard input for a focused BUFFER surface.
+ * `transition` means consume without mutating or replaying: the provider is
+ * absent, still opening, or mid-replacement.
+ */
+export type BufferOrdinaryInputOwnership =
+  | "provider"
+  | "transition"
+  | "none";
+
 export type BufferProviderRuntimeContext = {
   readonly workspaceRoot?: string;
   readonly agencHome?: string;
@@ -83,6 +93,7 @@ export class BufferProviderController {
   #lastOpen: BufferProviderOpenOptions | null = null;
   #lastSize: BufferProviderResize | null = null;
   #openGeneration = 0;
+  #openInFlight = false;
   #cleanupPromise: Promise<void> | null = null;
   #replacementPromise: Promise<boolean> | null = null;
   #restartPromise: Promise<boolean> | null = null;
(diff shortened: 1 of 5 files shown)


7etsuo (approved):
Reviewed on current main. Transition input is consumed across initial open, replacement, close, and cleanup without replaying into an unknown editor mode. Verification passed: focused BUFFER suites 91/91, both TypeScript configurations, production build, six PTY startup variants, and hosted affected tests.

PR #2044, opened 2026-09-02
test(memory): synchronize extraction shutdown on child-start barrier

### LINKED ISSUE #1939: test(memory): synchronize extraction shutdown coverage on a child-start barrier
## Problem

`runtime/tests/services/extractMemories/extractMemories.test.ts` proves that shutdown drains an active extraction, but it waits for the child to start with a fixed 20 iterations of `setTimeout(..., 0)`.

Under hosted shard load, those 20 scheduler turns can finish before path resolution and the memory-directory scan reach `runChild`. The assertion then reports that the child was never called, even though the test has not observed a product failure.

This happened on PR #1876 at exact head `9e1993452db6867fcb6a01f8037fd8d462d219fe`:

- `drains active extraction work before the caller finishes shutdown`
- failure at line 674: expected `runChild` once, received zero calls
- hosted run: https://github.com/tetsuo-ai/agenc-core/actions/runs/33444732386/job/99661120654

PR #1876 does not change `extractMemories.ts`, its tests, memory paths, or memory scanning.

## Root cause

The local `eventually` helper is a scheduler-spin count, not a synchronization contract. It assumes that 20 zero-delay timers are enough for unrelated async filesystem work to finish. That assumption changes with worker contention and filesystem latency.

The production implementation already registers the extraction promise before it begins the child and correctly exposes a drain function. The test needs to know that the mocked child has entered, not guess from elapsed scheduler turns.

## Suggested fix

Replace the polling assertion in this test with explicit two-phase deferred barriers owned by the `runChild` mock:

1. Create `childStarted` and `releaseChild` deferred promises.
2. Have `runChild` resolve `childStarted` immediately on entry, then await `releaseChild`.
3. Start `executeExtractMemories` and await `childStarted`. This is the exact precondition for calling `drainPendingExtraction`.
4. Start the drain and assert it remains unsettled while `releaseChild` is held.
5. Release the child, then await both extraction and drain.
6. Put release/cleanup in `finally` so a failed assertion cannot leave work running into another test.

Use the test framework's overall timeout only as a deadlock backstop with a clear failure message. Do not add sleeps, increase the polling count, weaken the assertion, or change the production drain timeout.

If several tests need this pattern, add a small typed deferred helper in test support and reuse it rather than introducing more local polling helpers.

## Regression verification

- Run this focused test repeatedly with shard-level worker contention.
- Prove it fails if the drainer stops waiting for registered work.
- Prove the drain remains pending until the child barrier is released.
- Confirm the test cleans up both barriers after an assertion failure.
- Run the full extraction test file and default-suite 3/4.

FalconOrtiz:
## Why

Hosted shard load can finish the local `eventually` helper's 20 `setTimeout(0)` turns before path resolution and the memory scan reach `runChild`. The drain shutdown test then fails with zero `runChild` calls even though production registered the extraction and would drain it. See #1939 and the flake on PR #1876 at `9e199345`.

## Scope

- `runtime/tests/services/extractMemories/extractMemories.test.ts`: replace the scheduler-spin wait in `drains active extraction work before the caller finishes shutdown` with `createControlledPromise` barriers (`childStarted` / `releaseChild`).
- Await child entry before `drainPendingExtraction`, assert the drain stays pending while the child is held, release in `finally` so a failed assertion cannot leak work.
- Reuse `tests/helpers/controlled-async.ts`. No production changes. No drain timeout change. No new sleeps or higher poll counts.

## Tradeoffs

Left the coalesce test on `eventually` for the smallest diff. Same race class exists there if hosted load worsens; barrierize it in a follow-up if it flakes.

## Blast Radius

Test-only. Consumers of memory extraction see no behavior change. Maintainers get a deterministic shutdown contract in the suite instead of a load-sensitive spin.

## Verification

- Forced repro: delayed `resolveMemoryDirectory` past 20 macrotasks → old `eventually` path failed with `expected runChild once, got 0`.
- Same delay with barriers → pass.
- Focused hermetic vitest `-t "drains active extraction work"` exit 0.
- Focused hermetic vitest `-t "extract memories service"` exit 0 (14 pass).
- Sonar self-check: unique New Code in one test; reused existing deferred helper; no `*Unlocked` re-indent clones; single `finally` site.

Closes #1939

### DIFF
--- a/runtime/tests/services/extractMemories/extractMemories.test.ts
+++ b/runtime/tests/services/extractMemories/extractMemories.test.ts
@@ -20,6 +20,7 @@ import {
 } from "./memory-paths.js";
 import { formatMemoryManifest, scanMemoryFiles } from "../../memory/index.js";
 import { resolveAgentRuntimeOptions } from "../../session/runtime-options.js";
+import { createControlledPromise } from "../../helpers/controlled-async.js";
 
 vi.mock("bun:bundle", () => ({ feature: () => false }));
 vi.mock("../../tools.js", () => ({}));
@@ -939,13 +940,17 @@ describe("extract memories service", () => {
   });
 
   it("drains active extraction work before the caller finishes shutdown", async () => {
-    let resolveChild!: () => void;
-    const runChild = vi.fn(
-      async () =>
-        new Promise<{ readonly outcome: "completed" }>((resolve) => {
-          resolveChild = () => resolve({ outcome: "completed" });
-        }),
+    const childStarted = createControlledPromise<void>(
+      "extractMemories child started",
     );
+    const releaseChild = createControlledPromise<void>(
+      "extractMemories child release",
+    );
+    const runChild = vi.fn(async () => {
+      childStarted.resolve(undefined);
+      await releaseChild.promise;
+      return { outcome: "completed" as const };
+    });
     initExtractMemories({
       env: {},
       minEligibleTurns: 1,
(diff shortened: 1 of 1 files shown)


FalconOrtiz:
**CI may be blocked on fork workflow approval** (same as sibling #2045). FalconOrtiz cannot approve (403). After workflows run, please confirm Sonar ship bar: **0.0% duplication**, 0 hotspots, coverage gate, 0 OPEN issues.

7etsuo (approved):
Reviewed on current main. The controlled start/release barrier removes polling from the shutdown assertion, the finally path prevents a stranded child, and the extraction-service suite passes (30/30).

PR #2052, opened 2026-09-03
fix(runtime): bind MCP content provider reads to verified handles

### LINKED ISSUE #1794: MCP content providers reopen validated paths
Verified on `main` at `065318d8f`.

## Problem

`readScopedRegularFile` in `runtime/src/mcp/server/content-providers.ts` validates a pathname with `lstat` and `realpath`, then reopens the canonical pathname with `readFile`. A writable workspace can replace the file or an ancestor between validation and the read. The helper also accepts multiply linked files.

This path reads skill prompts, memory, and instruction resources. Descriptor-bound readers elsewhere in the runtime already reject hard links and verify the opened file identity.

## Suggested fix

Replace the pathname reopen with a shared descriptor-bound reader. Open with `O_NOFOLLOW` and `O_NONBLOCK` where available, call `fstat`, require a regular file with one link, enforce a byte limit before allocation, and read from the verified handle. Confine traversal to an opened root descriptor when the platform supports it. On platforms without safe primitives, use an explicit fail-closed policy or a documented platform implementation rather than a silent weaker path.

## Acceptance criteria

- File and ancestor swaps between validation and open are rejected.
- Symlinks, hard links, FIFOs, devices, and oversized files are rejected.
- Tests use hooks to replace the candidate at each I/O boundary.
- Skill and resource reads share the same verified handle implementation.

signerless:
## Summary

- `readScopedRegularFile` validated a pathname with `lstat`/`realpath` and then reopened the canonical pathname with `readFile`, so a writable workspace could swap the file or an ancestor between validation and bytes; it also served multiply linked files (#1794).
- Skill prompts and resource bodies now share one descriptor-bound reader: validate, open once with `O_NOFOLLOW`/`O_NONBLOCK`, prove the opened object is the same single-link regular file validation saw, read it under a byte ceiling (`MAX_SCOPED_FILE_BYTES`, 1 MiB), and re-prove it after the read. File, ancestor, symlink, hard-link, FIFO, device, and oversized candidates are all rejected.
- Listing applies the same admission contract, so hard-linked or oversized candidates are never advertised as resources in the first place.
- The `readResourceContent` pathname seam is removed — there is no longer any route around the verified reader; deterministic race coverage swaps candidates at each I/O boundary via `beforeOpenForTesting`/`beforeReadForTesting` hooks, mirroring `secure-instruction-file.ts`.
- Node exposes no root-confined open (`openat2`/`RESOLVE_BENEATH`); containment remains canonical-resolution checks plus opened-handle identity proofs, documented in the module header like the runtime's other verified readers.

## Validation

- Reproduced on unchanged `main` end-to-end: `agenc mcp serve --transport stdio` in a workspace whose `.agenc/skills/leak/SKILL.md` was a hard link (nlink=2) to a private transcript outside the workspace — `prompts/list` served the skill and `prompts/get` returned the private body.
- Post-fix on the same live server: the hard-linked skill is rejected (`unknown prompt: leak`), a plain skill still serves, the hard-linked memory file is not listed, and `deploy-notes.md` serves with its token redacted.
- New `runtime/tests/mcp-server/content-providers-verified-reads.test.ts` (17 tests) swaps the candidate at each filesystem I/O boundary via the hooks: symlink, same-inode symlink, regular-file replacement, ancestor-directory swap, mid-read replacement, in-place mutation, hard link, oversized, FIFO. 14/17 fail with the source change reverted (the 3 passing cases are happy-path controls).
- Reworked `runtime/tests/mcp-server/prompts-resources.test.ts` to write redaction-corpus fixtures to real files instead of injecting through the removed seam: 33/33 across both provider test files.
- `npm run typecheck` passed.
- `npm run test:fast -- --base origin/main` passed: 41 files, 1032 tests.

Closes #1794

### DIFF
--- a/runtime/src/mcp/server/content-providers.ts
+++ b/runtime/src/mcp/server/content-providers.ts
@@ -18,9 +18,21 @@
  *     client-supplied paths. Only the selected resource body is read,
  *     then its contents pass through the canonical egress secret sanitizer.
  *
+ * Skill and resource bodies are served by one shared descriptor-bound
+ * reader (`readScopedRegularFile`): a candidate is validated, opened once
+ * with `O_NOFOLLOW`/`O_NONBLOCK`, proven to be the same single-link
+ * regular object that validation saw, read from that handle under a byte
+ * ceiling, and re-proved after the read. A pathname is never reopened
+ * after validation, so a writable workspace cannot swap the file or an
+ * ancestor between checks and bytes. Node exposes no root-confined
+ * open (openat2/`RESOLVE_BENEATH`); as in the runtime's other verified
+ * readers, containment is enforced on the canonical resolution plus
+ * opened-handle identity proofs rather than a silent weaker path.
+ *
  * @module
  */
-import { lstat, readdir, readFile, realpath } from "node:fs/promises";
+import { constants as fsConstants, type BigIntStats } from "node:fs";
+import { lstat, open, readdir, realpath } from "node:fs/promises";
 import { basename, isAbsolute, join, relative } from "node:path";
 
 import {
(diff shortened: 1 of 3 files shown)


signerless:
Post-merge note, for anyone reading this PR as the resolution of #1794: the
ancestor-swap escape is still reachable on `main` after `f75f0a2ac`, so #1794
is reopened. Reproduction, numbers and the mechanism are in
https://github.com/tetsuo-ai/agenc-core/issues/1794#issuecomment-5530644066
rather than repeated here.

Short version: containment is still decided twice by resolving a pathname
(`content-providers.ts:200` and `:247`), and neither resolution is tied to the
descriptor that produced the bytes. Flipping an ancestor between the validating
`lstat` and the `realpath` yields out-of-scope bytes under an in-scope
`canonicalPath` — 10 forged / 30,033 attempts seam-free, 12 / 42,394 through
`getPrompt`, where the payload is the whole skill body.

What this PR did fix stands: wholesale file swap, hard links, the FIFO hang,
the size ceiling, and the open→fstat and read-loop windows.

Worth flagging separately for future PRs on this file: five of the guards here
are unpinned. Deleting the `scopeRoot` containment check, `O_NOFOLLOW`, the
`isFile()` clause, or either after-read re-check leaves all 33 tests green, so
a green suite did not distinguish this change from one without them.

A descriptor-bound fix is in progress as a separate PR against `main`.

PR #2155, opened 2026-09-04
feat(runtime): add native Kimi provider

signerless:
## Summary

Add native global Moonshot/Kimi support with its own `MOONSHOT_API_KEY` and fixed official endpoint. Expose the harness's normal tools without the reduced local-compatible profile, while retaining all tool permissions.

- Curate the four models currently returned by the official API: K3, K2.7 Code, K2.7 Code Highspeed, and K2.6.
- Match documented context, output, reasoning, image input, function-calling, and streaming contracts.
- Preserve model-scoped reasoning across tool continuations and subsequent turns, including after compaction.
- Add provider-specific pricing and cache-token accounting.
- Add a headless `kimi-models --json` command for bounded account-authoritative discovery from Desktop through the normal Core transport.
- Keep media generation provider-independent: a Moonshot chat key does not create an undocumented image-generation backend.

## Verification

- Packaged Desktop E2E: K3 performed 7 tool calls, created two JavaScript files, ran 8/8 Node tests, discovered and called desktop browser MCP, read the actual PNG pixels, and rendered the image inline. Turn completed normally with the daemon still connected.
- Real API authentication, balance, and model discovery passed.
- All four models completed tool-call → tool-result → final-answer loops.
- K3 completed a further user turn with preserved reasoning.
- K3 and K2.7 Code interpreted actual base64 pixels returned by a tool (red square, blue circle, and text `KIMI 742`).
- K2.7 Code completed a structured-output + tools loop. K3 did not honor required tools when combined with JSON Schema, so that combination remains disabled.
- Highspeed returned a transient upstream overload once, then passed on a bounded retry.
- No credential is committed. Live tests use a temporary environment variable and synthetic data.

## Official references

- https://platform.kimi.ai/docs/api/models-overview
- https://platform.kimi.ai/docs/guide/use-thinking-models
- https://platform.kimi.ai/docs/guide/use-kimi-vision-model
- https://platform.kimi.ai/docs/guide/use-tool-choice
- https://platform.kimi.ai/docs/api/chat


7etsuo (changes requested):
Request changes at `20a7758f8029efdba5e112838de371f04964ef5c`.

[P2] Keep the unrelated skill-candidate home behavior from main when resolving the conflict in `runtime/src/services/extractMemories/extractMemories.ts`. This PR removes the `resolveSkillCandidatesHome` fallback and changes the test to disable proposals even when the injected environment explicitly names `AGENC_HOME`. Main's #2218 implementation routes that fallback through `resolveHomeContext`, preserving named injected homes while rejecting default-home fallback and retired variables. A native provider addition should not silently disable that supported extraction path. Drop this unrelated hunk or separate an intentional contract change with its callers addressed.

The provider work also needs a rebase and broader validation: shared OpenAI-compatible streaming, message normalization, reasoning replay, capability gating, and metadata code change for multiple providers. GitNexus marks the shared factory and wire builders critical because of their callers. Current green checks are for the old base and do not establish compatibility with today's main.

Runtime typecheck and the Kimi provider/CLI, Z.AI, provider-parity, and cost suites passed on this head. I checked the full diff, including credential isolation, endpoint validation, malformed/unterminated SSE refusal, reasoning provenance, tool limits, image restrictions, and usage accounting. Rebase, preserve the home contract, then run the full provider/wire suites and repository full-suite/platform gates. No authenticated live Kimi request was made during this review.

PR #2248, opened 2026-09-07
workflow: a reviewer that never answered is a known, retried failure

signerless:
## Why

Desktop soak `goal-14` (the retest of #2245 on a daemon built from main at 4e92c3f6a): plan, implement and verify passed, then the review stage died within 250 ms. The reviewer's single model call returned HTTP 403 `The OAuth2 access token could not be validated.` (provider trace `agent-logs/review-2209…/llm-00001.jsonl`) while the verify child's calls had succeeded one second earlier on the same daemon and a prompt through the same daemon worked again eight minutes later. The seam threw the provider error as a plain `Error`; the controller's catch around `runIndependentReview` knew only `ReviewParseError`, so `#driveEffect` classified the dispatched spawn as `spawn_failed_after_dispatch_without_acknowledgement` and the run ended `unknown_outcome` with an operator review pending, at $8.25.

A read-only reviewer with no tools attached that settled without a response has committed nothing, so the honest classification is a known failure with the stage's bounded retry, the same path an unparseable review already takes.

## What changed

- `runtime/src/workflow/independent-review.ts`: `ReviewInvocationError` (message `independent review produced no output: <detail>`, keeps the cause).
- `runtime/src/app-server/workflow/session-adapters.ts`: the reviewer seam throws `ReviewInvocationError` when the one-shot review ends with no text, wrapping `outcome.error` when there is one and the verdict otherwise, instead of rethrowing the raw provider error.
- `runtime/src/app-server/workflow/verified-change-controller.ts`: the review stage catches `ReviewInvocationError` alongside `ReviewParseError`, records a failed review child terminal that names the cause and returns a failed attempt with `failure.reason = "review_invocation_failed"` (`review_unparseable` stays for the parse case). `#runStageWithRetries` then reruns the reviewer once; a second failure ends the run `failed / step_retries_exhausted`.
- `runtime/tests/workflow/verified-change-controller.test.ts`: `FakeReviewer` gains an `errors` queue; two cases: one failed invocation retries once and the run completes with the first attempt durable as a failed child terminal naming the 403; two failed invocations end the run `failed`, and no effect of the run is `unknown_outcome`.

## Evidence

| run | result |
| --- | --- |
| `vitest run tests/workflow/verified-change-controller.test.ts tests/workflow/session-adapters.policy.test.ts` | 44 passed (42 before, plus the two new cases) |
| `tsc --noEmit`, `tsc --noEmit --project tsconfig.test-support.json` | no errors outside this worktree's `TS2883` lodash baseline |
| goal-14 rollouts | reviewer run `review-2209…` terminal `failed`, stop reason `grok authentication failed (HTTP 403)`; parent run terminal `unknown_outcome`, `workflow.review failed after dispatch without acknowledgement` |

## Tests

The new cases drive the real controller through the M5 harness with a reviewer that throws `ReviewInvocationError` on its first call (and on both calls in the second case) and assert the effect outcomes, the child terminal and the run terminal.

## Not changed

- Why the bearer was stale: the adapter's only recovery for admitted single-wire-attempt calls is the pre-flight `refreshOAuthBearerIfExpiring`, which refreshes when the stored grant is expiring but does not re-read a bearer that another provider instance already refreshed. That is a separate fix; this PR makes the failure survivable.
- Errors other than a settled reviewer without output keep the unknown-outcome path.

## Counterparts

Soak finding F76; follows #2245 (goal-13 F73). The auth-side issue is filed separately.

### DIFF
--- a/runtime/src/app-server/workflow/session-adapters.ts
+++ b/runtime/src/app-server/workflow/session-adapters.ts
@@ -62,7 +62,10 @@ import { runSupervisedProcess } from "../../utils/supervisedProcess.js";
 import { applyUnattendedPermissionPolicyToContext } from "../../permissions/unattended-policy.js";
 import type { PermissionModeRegistry } from "../../permissions/permission-mode.js";
 import type { StateRunDurabilityRepository } from "../../state/run-durability.js";
-import type { ReviewerInvoker } from "../../workflow/independent-review.js";
+import {
+  ReviewInvocationError,
+  type ReviewerInvoker,
+} from "../../workflow/independent-review.js";
 import type {
   WorkflowCommandResult,
   WorkflowCommandRunner,
@@ -977,11 +980,13 @@ export function createWorkflowSessionSeams(
         },
       );
       if (outcome.rawText === null) {
-        throw (
-          outcome.error ??
-          new WorkflowSessionSeamError(
-            `independent review produced no output (${outcome.verdict})`,
-          )
+        // The one-shot settled without a response. Typed so the controller
+        // treats it as a known, retryable failure (soak F76).
+        throw new ReviewInvocationError(
+          outcome.error !== undefined
+            ? errorMessage(outcome.error)
+            : `verdict ${outcome.verdict}`,
+          outcome.error !== undefined ? { cause: outcome.error } : undefined,
         );
       }
       return outcome.rawText;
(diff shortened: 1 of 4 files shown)


7etsuo (commented):
Completed the requested review of `a928e07667cf9b699b858712b92ae86d8f1ff254`; this PR was merged by another session while the review was in progress.

No blocking finding. The session adapter converts a settled no-output invocation into the typed failure, the controller records a durable failed review child, and retry remains bounded. Unclassified failures still follow the existing unknown-outcome path. Runtime typecheck and the controller, session-policy, independent-review, and repair suites passed locally.

PR #2249, opened 2026-09-07
llm: the stored xAI OAuth grant wins in the grok factory itself

signerless:
## Why

Desktop soak `goal-14`: the verified-change reviewer's single Grok call got HTTP 403 `The OAuth2 access token could not be validated.` one second after the verify child's calls had succeeded on the same daemon, and a prompt through the same daemon worked again minutes later. The review delegate builds its provider with `createProvider(providerName, { ...readProviderFactoryOptions(parentProvider), model })`. The recorded options hold the bearer the parent resolved when it was created (`markFactoryProvider` records `apiKey`), and `readProviderFactoryOptions` returns them unchanged. The parent here was the run session's provider, created at 06:58:41 and never used for a model call; by 07:22:44 sibling child providers had refreshed the stored grant, so the snapshot no longer matched it, `isXaiOauthBearer` said no, and the factory built the reviewer's provider in API-key mode: no refresh callbacks, so no pre-flight refresh, and a dead bearer on the wire with no recovery.

The product rule already says the stored grant wins over any passed key ("/grok-login OAuth wins over an explicit env-style apiKey"), but only `resolveProviderFactoryOptions` applied it, and twelve call sites re-create providers from recorded options without going through it (review delegate, compaction, memory selector, code prediction, admitted model calls, bootstrap services, model-facing tools).

## What changed

- `runtime/src/llm/provider.ts` (grok branch): when `credentialHome` is given, read the stored grant with `readXaiOauthAccessToken`; if a live one exists the factory is in OAuth mode and uses its current bearer, whatever `apiKey` was passed. Without a grant the previous behaviour stands (passed key, or the "requires apiKey" error).
- `runtime/tests/llm/grok-oauth-factory.test.ts`: the harness also exposes the raw factory; two cases: a stale bearer snapshot passed as `apiKey` yields the stored bearer and OAuth mode (a non-xAI base URL is refused); a provider re-created from the recorded options of one built before a refresh carries the current bearer with the refresh callbacks installed.

## Evidence

| run | result |
| --- | --- |
| `vitest run tests/llm/grok-oauth-factory.test.ts tests/llm/providers/grok/adapter.test.ts tests/llm/provider-credential-authority.test.ts` | 66 passed (64 before, plus the two new cases) |
| `vitest run tests/utils/xaiOauthCredentials.test.ts` (covers `isXaiOauthBearer`, which stays although the factory no longer calls it) | passed |
| `tsc --noEmit`, `tsc --noEmit --project tsconfig.test-support.json` | no errors outside this worktree's `TS2883` lodash baseline |
| goal-14 provider trace `agent-logs/review-2209…/llm-00001.jsonl` | one request, `error` after 250 ms, status 403, body `The OAuth2 access token could not be validated.`; the verify child's trace ends with a successful response at 07:22:43 |

## Tests

The new cases drive the real `createProvider` with the credentials module mocked, as the existing tests in the file do; the second case reproduces the failing sequence exactly: build, refresh the stored grant, re-create from recorded options.

## Not changed

- The option resolver's own OAuth substitution and the API-key mode without a grant (its base-URL exemption test still passes).
- The recorded factory options still carry `apiKey`; with the factory re-resolving the grant the snapshot is harmless for grok.
- `isXaiOauthBearer` stays in the credentials module (its unit test covers it); the factory no longer calls it.
- The controller side of the same incident is #2248 (a reviewer that never answered is a retried known failure).

## Counterparts

Soak finding F76; #2248.

### DIFF
--- a/runtime/src/llm/provider.ts
+++ b/runtime/src/llm/provider.ts
@@ -80,7 +80,7 @@ import {
 export { resolveBuiltInProviderSlug } from "./registry/provider-info.js";
 import {
   forceRefreshXaiOauthCredentials,
-  isXaiOauthBearer,
+  readXaiOauthAccessToken,
   xaiOauthRequiresRelogin,
 } from "../utils/xaiOauthCredentials.js";
 import { isTrustedXaiOauthInferenceBaseUrl } from "../services/xai/oauth.js";
@@ -1614,10 +1614,19 @@ export function createProvider(
       // X means subscription access; leftover XAI_API_KEY must not shadow it.
       // Bearer refreshes via the adapter's I-14 401-recovery hook.
       const factoryApiKey = resolveFactoryApiKey(opts);
-      const usesXaiOauth =
-        opts.credentialHome !== undefined &&
-        isXaiOauthBearer(opts.credentialHome, factoryApiKey);
-      const apiKey = factoryApiKey ?? requireFactoryApiKey("grok", opts);
+      // The stored grant wins whenever it exists, here and not only in the
+      // option resolver upstream. Soak F76: a provider re-created from
+      // another instance's recorded factory options carries that instance's
+      // bearer snapshot; once the stored grant has been refreshed the
+      // snapshot no longer matches, and treating it as an API key sends a
+      // dead token with no refresh path (xAI answers 403).
+      const storedOauthBearer =
+        opts.credentialHome !== undefined
+          ? readXaiOauthAccessToken(opts.credentialHome)
+          : undefined;
+      const usesXaiOauth = storedOauthBearer !== undefined;
+      const apiKey =
+        storedOauthBearer ?? factoryApiKey ?? requireFactoryApiKey("grok", opts);
       const model = requireModel("grok", opts.model, defaultModelFor("grok"));
       const cfg: GrokProviderConfig = {
         ...buildCommonConfig(extra),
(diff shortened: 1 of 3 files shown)


7etsuo (approved):
Reviewed the updated head `15af2c9a7bbd169bd55d1d250ce747e034421efe`. No blocking finding.

The raw Grok factory now gives the stored, non-quarantined OAuth grant the same precedence as the option resolver. Re-created providers receive the current bearer and refresh callbacks, and the trusted-endpoint check still runs before use. Without a stored grant, the existing API-key behavior remains. The new stale-snapshot and factory-option replay cases pass.

The latest commit restores `isXaiOauthBearer`; its existing credential tests pass as well. Validation and approval refer to this updated head.

Validation: the three candidate PRs were combined on current main in integration commit `8587f0e045ba11e031ce45bbe49d06abb797713d`. The combined tree passes 285 focused tests, runtime typecheck, build, PTY startup, and all nine native platform jobs. The restored OAuth credential helper also passes its own suite.

The complete hosted default suite has three green shards. Shard 3 has one failed benchmark binding for `runtime/package.json` and four Darwin-only skips under Linux; both are also present on unchanged main, and the relevant files are identical between main and this integration. No new failure remains in the candidate tree. This is not a claim that main's full suite is green.

Evidence: [final candidate validation](https://github.com/tetsuo-ai/agenc-core/actions/runs/34098153762), [unchanged main baseline](https://github.com/tetsuo-ai/agenc-core/actions/runs/34096518321).

The narrower PR affected-tests run is still in progress; the completed full-suite, typecheck, and platform validation above cover this exact head in the candidate integration. GitHub reports no required check for this branch.

PR #2310, opened 2026-09-08
fix(llm): coordinate OAuth refresh by credential generation

### LINKED ISSUE #2060: fix(llm): single-flight OpenAI OAuth refresh and ignore stale auth failures
## Summary

`OpenAIAuthSession.withAuthorizedOperation()` shares mutable OAuth state and an exhausted marker across operations, but refreshes are not coordinated. Concurrent operations that receive an auth failure can refresh the same token twice; a slower failed refresh can then overwrite a newer successful state and poison later calls.

Verified against `origin/main` at dbe3c1f4e7c120f72f86b7c776939590990b70f9.

## Evidence

- [shared OAuth state and exhaustion marker](https://github.com/tetsuo-ai/agenc-core/blob/dbe3c1f4e7c120f72f86b7c776939590990b70f9/runtime/src/llm/providers/openai/auth.ts#L22-L25)
- [authorize/retry/refresh flow](https://github.com/tetsuo-ai/agenc-core/blob/dbe3c1f4e7c120f72f86b7c776939590990b70f9/runtime/src/llm/providers/openai/auth.ts#L46-L81)
- [multiple public adapter operations use the shared session](https://github.com/tetsuo-ai/agenc-core/blob/dbe3c1f4e7c120f72f86b7c776939590990b70f9/runtime/src/llm/providers/openai/adapter.ts#L580-L810)

A controlled two-operation reproduction produced two refresh calls. One operation completed with the new credentials while the delayed refresh failed; that stale failure set `oauthExhaustedMessage`, and the next operation failed immediately without attempting the provider call.

Normal model turns may serialize elsewhere, but the provider exposes other overlapping operations (health/model discovery, stored-response work, embeddings, and direct adapter consumers), so the shared object still needs internal coordination.

## Impact

- Rotating or one-time refresh credentials can be consumed twice.
- A valid refreshed session can be replaced by stale failure state.
- Subsequent requests can be rejected until the provider/session is reconstructed.

## Proposed direction

Make refresh single-flight per auth generation. Waiters should reuse the in-flight refresh result. Only a refresh that still corresponds to the current credential generation may update state or mark authentication exhausted. Define cancellation so cancelling one waiter does not cancel refresh work still needed by others.

## Acceptance criteria

- Simultaneous 401/auth failures trigger one refresh.
- A delayed failure from an older generation cannot overwrite newer valid credentials.
- Genuine refresh exhaustion remains sticky for that generation.
- Non-auth failures retain existing retry behavior.
- Cancelling one waiter does not incorrectly cancel or poison other waiters.
- Deterministic tests cover the successful/stale-failure interleaving.

7etsuo:
## Changes

- Coalesce concurrent OpenAI OAuth failures into one refresh per credential generation. Bind request headers to that generation across asynchronous setup.
- Reuse new credentials after delayed old-request failures. Keep genuine exhaustion sticky, including when an outstanding request fails after exhaustion.
- Forward chat and stream cancellation to each refresh waiter without cancelling shared work. Observe late rejection and remove abort listeners.
- Preserve single-wire no-refresh behavior, non-auth errors and bounded retries. Add deterministic race and adapter cancellation tests and document the behavior.

## Validation

- Original implementation fails all four initial regression tests. A further interleaving test caught and now prevents revival after genuine exhaustion.
- Docker-isolated targeted tests and runtime/test-support typechecks pass: 74 tests across four files, zero skips.
- Bun 1.3.12 passes all 20 new race tests.
- The supported Linux native slice passes 92 tests across five files.
- Final build, generated SDK contract and all PTY startup checks pass.
- Full root `npm test` passes on 84be9d3d54032079ec2016648589dc82f1a649db: 23,549 runtime tests across 2,278 files, zero failures or skips, plus the root and launcher checks.
- GitNexus identifies this as a critical authentication boundary with five direct adapter callers. The staged diff contains only the expected auth, adapter, tests and documentation changes. No changed production module is in the FND measured closure.
- Hosted test workflows remain disabled. No Actions runs are requested. Native Darwin and Windows execution is not claimed.

Fixes #2060.

---

> [!NOTE]
> **High Risk**
> Changes sit on the OpenAI OAuth authentication boundary with complex concurrency; incorrect refresh coalescing or exhaustion handling could break sign-in sessions or allow stale credentials to win.
> 
> **Overview**
> **OpenAI OAuth refresh is reworked** so concurrent 401s on the same credential generation share one in-flight refresh, while each request keeps the bearer token it started with via `AsyncLocalStorage`.
> 
> `OpenAIAuthSession` replaces the generic `retryWithOAuthRefresh` loop with generation-aware recovery: stale operations cannot mark a newer session exhausted, genuine exhaustion stays sticky until re-login, and `singleWireAttempt` still skips refresh entirely. Chat/stream now pass `AbortSignal` into `withAuthorizedOperation`, so cancelling one caller aborts only that waiter’s wait on refresh (listeners cleaned up) without stopping the shared refresh. The OpenAI adapter wires `options.signal` through for chat and stream.
> 
> Provider docs describe the concurrency, cancellation, and exhaustion rules. A large `auth-concurrency.test` suite covers coalescing, races, cancellation, and adapter-level chat/stream behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84be9d3d54032079ec2016648589dc82f1a649db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/reference/providers.md
+++ b/docs/reference/providers.md
@@ -192,6 +192,12 @@ stores credential values.
   retired native `agenc` credential field are explicit one-way migration
   inputs only. Reads, refreshes, and clears stay bound to the client's captured
   `HomeContext`, and refresh compare-and-swap preserves a newer login.
+  Within an OpenAI provider instance, concurrent 401 failures share one
+  refresh for the credentials used by those requests. Delayed failures from
+  older credentials cannot invalidate a refreshed session. Cancelling a chat
+  or stream stops that caller's wait without cancelling the shared refresh.
+  Genuine refresh exhaustion stays in effect until the provider is recreated
+  after login. Single-wire requests never refresh or mark refresh exhausted.
   List reachable models with `agenc openai-models --json`
   (`{ok, models, authMode}`; tokens never in the output). See
   [cli.md](cli.md#openai-models).
--- a/runtime/src/llm/providers/openai/adapter.ts
+++ b/runtime/src/llm/providers/openai/adapter.ts
@@ -755,7 +755,7 @@ export class OpenAIProvider implements LLMProvider {
           maxTokenField: this.resolveChatCompletionsMaxTokenField(),
           providerCapabilityHints,
         });
-      }, { singleWireAttempt: options?.singleWireAttempt });
+      }, { singleWireAttempt: options?.singleWireAttempt, signal: options?.signal });
     } catch (error) {
       if (isFallbackTriggeredError(error)) {
         throw error;
@@ -801,7 +801,7 @@ export class OpenAIProvider implements LLMProvider {
           timeoutMs,
           headers,
         );
-      }, { singleWireAttempt: options?.singleWireAttempt });
+      }, { singleWireAttempt: options?.singleWireAttempt, signal: options?.signal });
     } catch (error) {
       if (isFallbackTriggeredError(error)) {
         throw error;
--- a/runtime/src/llm/providers/openai/auth.ts
+++ b/runtime/src/llm/providers/openai/auth.ts
@@ -6,12 +6,15 @@
  * @module
  */
 
+import { AsyncLocalStorage } from "node:async_hooks";
 import {
   assertNonEmptyApiKey,
   buildBearerAuthHeaders,
 } from "../../auth/bearer.js";
 import {
-  retryWithOAuthRefresh,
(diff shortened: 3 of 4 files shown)


7etsuo:
cursor review

PR #2325, opened 2026-09-08
fix(compact): remove microcompaction's dead time-window exemption

### LINKED ISSUE #2253: Microcompaction's recent-result exemption is dead: every converted message is stamped with the epoch
## What

Microcompaction has a time-window exemption that never fires in the live pipeline, so its documented protection against clearing a hot tool result is not in force.

`isWithinTimeWindow` (`runtime/src/services/compact/microCompact.ts`) reads `message.timestamp`:

```ts
function isWithinTimeWindow(message: RuntimeMessage, now: number, clearAfterMs: number): boolean {
  if (!message.timestamp) return false;
  const timestamp = Date.parse(message.timestamp);
  return Number.isFinite(timestamp) && now - timestamp < clearAfterMs;
}
```

Its result becomes `recent`, one of the three victim exemptions (`if (victim.recent || isLatestReadOfPath) ...`).

But `runtime/src/session/runtime-message-conversion.ts` stamps every converted message with the epoch:

```ts
timestamp: new Date(0).toISOString(),
```

in both conversion paths. So `now - timestamp` is about 1.77e12 ms and the comparison against the default 5-minute window is always false. `recent` is always false, and no result is ever protected for being fresh. The other two exemptions, the keep-recent-N window and latest-read-per-path, still work, so the effect is a missing guard rather than a broken valve, but it means a result produced seconds ago can be cleared the moment it leaves the flat recent-5 window.

## Where it came from

`RuntimeMessage.timestamp` is optional in `runtime/src/services/compact/types.ts`, and nothing on the live path synthesizes a real one. `LLMMessage` (`runtime/src/llm/types.ts`) has no timestamp field at all, so the conversion has nothing to carry across and writes a constant.

## Options

1. Carry a real time. Give the conversion the message's actual arrival time, which means adding a field to the live message shape or a side table keyed by tool-call id.
2. Drop the exemption. If nothing can supply a time, delete `isWithinTimeWindow`, the `recent` field and the config knob rather than keeping a guard that reads as protection and is inert.

Either is fine; the current state is the worst one, because the code and the config both suggest a protection that does not exist.

## How it surfaced

Found while tracing the prompt-cache prefix losses in #2244 (desktop soak `goal-15`). Not the cause of those, and not fixed in #2252, which changes the in-memory bound only.

signerless:
Closes #2253.

## Why

Microcompaction exempted a tool result from clearing while it sat inside a time window, sized by `AGENC_MICROCOMPACT_CLEAR_AFTER_MS`. The exemption could never fire.

`isWithinTimeWindow` read `message.timestamp`. The only producer of these messages is `toAgenCRuntimeMessages`, which stamps `new Date(0)` in both of its paths, so `now - timestamp` was always the age of Unix time and `recent` was always `false`.

**That constant is not the bug, and it stays.** One history has to project to the same bytes on every call, which is the property the batched valve exists to protect: a wall-clock reading there would break the provider's cached prefix on every projection instead of once per batch. The constant is deliberate, and it is precisely why a time window can never work on this path.

Two of the three victim exemptions do work (keep-recent-N and latest-read-per-path), so this was a missing guard rather than a broken valve. But a guard that cannot run, paired with an environment variable that reads as a control over it, documents a protection nobody has. Anyone who set that variable got nothing.

Wall-clock age is also the wrong measure here. Resume an hour-old conversation and every result looks old, including the one the model is about to use; run a long turn and stale results stay young. Position is the useful measure, and `MICROCOMPACT_KEEP_RECENT` already covers it.

## What changed per file

- **`runtime/src/services/compact/microCompact.ts`** — drops `isWithinTimeWindow`, the `recent` field on `CompactableResultPosition`, and the `now`/`clearAfterMs` plumbing through `collectCompactableToolResultPositions`. The valve doc comment no longer promises a time window.
- **`runtime/src/services/compact/timeBasedMCConfig.ts`** — deleted. Its only consumer was the line above.
- **`runtime/src/session/runtime-message-conversion.ts`** — no behaviour change. Both stamp sites gain a note saying the constant is deliberate and that nothing may derive recency from it, so this is not rebuilt on the same footing.
- **`docs/reference/env.md`**, **`runtime/knip.config.mjs`** — `AGENC_MICROCOMPACT_CLEAR_AFTER_MS` and the deleted module drop out.
- **`runtime/tests/services/compact/compact-surfaces.test.ts`** — the knob's two assertions go; the snip-compact half of that test stays.
- **`runtime/tests/services/compact/retired-microcompact-time-window.architecture.test.ts`** — new, following the repo's `retired-*` pattern.

## Evidence

The retirement test does not just grep. It converts three messages through the real producer and parses each stamp:

```
expect(Date.parse(message.timestamp!)).toBe(0);
```

All three are the epoch, which is the whole claim: the subtraction never landed inside any window an operator could set. Removing the exemption is therefore behaviour-preserving by construction, not by argument.

## Tests

`runtime/tests/services/compact/` plus the three env-documentation architecture suites: **22 files, 153 passed**. Typecheck clean over the changed files (the repo's TS2883 lodash `MemoizedFunction` findings are the existing baseline and are untouched).

`npm run check:unused` fails on this branch and on `main` alike (41 unused files, 771 unused exports); knip is not wired into any workflow, so that baseline is unrelated to this change.

## Not changed

- The other two victim exemptions, the pressure limit, and the batching behaviour are untouched.
- `RuntimeMessage.timestamp` stays optional in `types.ts`. Nothing in `services/compact/` reads it now, but the field is part of the wire shape.
- Option 1 in the issue (carry a real arrival time) is not taken: there is nowhere honest to get one, since `LLMMessage` has no time field and the projection must stay deterministic.


signerless:
Verified locally, because GitHub Actions has produced no runs on this repository since 2026-09-08T00:00:00Z (Actions is enabled; the stall covers every open PR, not this branch). SonarCloud is the only reporting check here and both of its checks pass.

`tests/services/compact/`, `tests/session/`, `tests/config/` on this branch: **160 files, 2437 passed, 6 failed**.

The 6 are all in `tests/config/session-home-authority.test.ts` and are **not from this change**. Checked out clean `origin/main` (a426fcee9) in a separate worktree and ran the same file: the same 6 fail there. They are sensitive to the `AGENC_HOME` my isolated workspace sets, and nothing in this diff touches home or config resolution.

PR #2329, opened 2026-09-09
fix(runtime): enforce JSON line byte limits before dispatch

### LINKED ISSUE #2073: fix(runtime): enforce JSON-line caps at the terminating newline
## Summary

The app-server and MCP stdio transports reject newline-free input after it exceeds the configured line cap, but both lose the accumulated byte count when the chunk that crosses the cap also contains the terminating newline. The oversized completed line is then parsed and dispatched.

Verified against `origin/main` at dbe3c1f4e7c120f72f86b7c776939590990b70f9. This narrows the original finding: PR #910 fixed truly unbounded newline-free growth; the remaining defect is a bounded delimiter-boundary violation.

## Evidence

- [app-server byte counter](https://github.com/tetsuo-ai/agenc-core/blob/dbe3c1f4e7c120f72f86b7c776939590990b70f9/runtime/src/app-server/transport/stdio.ts#L81-L101)
- [app-server line dispatch](https://github.com/tetsuo-ai/agenc-core/blob/dbe3c1f4e7c120f72f86b7c776939590990b70f9/runtime/src/app-server/transport/stdio.ts#L105-L106)
- [MCP byte counter](https://github.com/tetsuo-ai/agenc-core/blob/dbe3c1f4e7c120f72f86b7c776939590990b70f9/runtime/src/mcp-server/stdio.ts#L84-L103)
- [MCP line dispatch](https://github.com/tetsuo-ai/agenc-core/blob/dbe3c1f4e7c120f72f86b7c776939590990b70f9/runtime/src/mcp-server/stdio.ts#L107-L109)
- [production Unix-socket use](https://github.com/tetsuo-ai/agenc-core/blob/dbe3c1f4e7c120f72f86b7c776939590990b70f9/runtime/src/app-server/transport/unix-socket.ts#L333-L336)
- [MCP server use](https://github.com/tetsuo-ai/agenc-core/blob/dbe3c1f4e7c120f72f86b7c776939590990b70f9/runtime/src/mcp/server/start.ts#L182-L197)

Both counters do this:

```ts
if (lastNewline === -1) {
  unterminatedBytes += data.length;
} else {
  unterminatedBytes = data.length - lastNewline - 1;
}
```

If an earlier chunk leaves the line just below the limit and the next chunk crosses it and includes `\n`, the `else` branch replaces the total with only the bytes after the last newline (usually zero). Because `readline` installed its own input listener first, checking only in this later `data` listener can also occur after line dispatch.

## Reproduction

With `maxLineBytes: 64`, a single newline-terminated JSON frame of 190 bytes was delivered to the app-server `onMessage` callback with no error. The equivalent MCP probe processed a 198-byte line.

A real Node 26 Unix-socket probe using the default limit observed:

```json
{
  "frameBytes": 16777218,
  "cap": 16777216,
  "maxServerDataChunk": 65536,
  "messages": 1,
  "errors": []
}
```

The control case (65 bytes without a newline and a 64-byte cap) correctly destroyed the input and delivered no message.

## Impact

A malformed or buggy local client can make the runtime allocate, parse, and dispatch a JSON frame larger than the advertised shared 16 MiB envelope. Production overshoot was bounded by the crossing socket chunk; generic `Readable` users can supply larger chunks.

## Proposed direction

Use one bounded JSON-line framer shared by app-server and MCP. It must carry the previous segment length into bytes before every newline and validate a completed line before exposing it to parse/dispatch. A transform/custom bounded decoder is safer than a second listener attached after `readline`.

## Acceptance criteria

- A line exactly `maxLineBytes` bytes is accepted.
- A line at `maxLineBytes + 1` is rejected with the newline in the same chunk or a later chunk.
- A crossing chunk containing additional valid lines cannot dispatch the oversized line.
- Existing newline-free overflow behavior remains intact.
- Multiple individually bounded lines in one chunk remain valid.
- CRLF and split multibyte UTF-8 input preserve current behavior.
- Both transports use the same tested framing invariant.

Related but not duplicate: PR #910 covers only newline-free growth. No existing issue or open PR covers the terminating-newline case.

7etsuo:
## Changes

- Share one bounded input reader between daemon and MCP stdio transports. Count every line's UTF-8 bytes before decoding or dispatch, including the segment before a terminating newline.
- Keep exact-limit lines, CRLF, bare CR, split UTF-8, multiple lines per chunk, and the final bounded line at EOF. Preserve transport dispatch ordering and pending-work drain.
- Close oversized input once, discard the unfinished frame, remove listeners, and destroy input even when an error callback throws.
- Keep buffered bytes within the configured cap and copy trailing fragments without retaining the original chunk. Document the 16 MiB payload limit.

## Validation

- Fourteen regression cases fail on the original transports. The real Unix-socket case also rejects the original implementation at 16 MiB plus one byte, while its exact-limit control passes.
- Both TypeScript gates pass. The focused suite passes 103 tests across 7 files, and the Linux native checks pass 92 tests across 5 files.
- Build and real-PTY startup pass. The full local runtime suite passes all 23,941 tests across 2,297 files.
- Hosted test CI stays disabled. No workflow was enabled, dispatched, rerun, or retried.

Closes #2073.

### DIFF
--- a/docs/reference/daemon.md
+++ b/docs/reference/daemon.md
@@ -135,6 +135,12 @@ const client = await connect(); // socket + cookie under AGENC_HOME
 
 ## Protocol
 
+The daemon's local socket and the MCP stdio server accept at most 16 MiB of
+UTF-8 payload per JSON line, excluding the LF, CRLF, or CR delimiter. A line
+exactly at the limit is valid. An oversized line closes the input before JSON
+parsing or dispatch, including when the terminating newline arrives in the
+chunk that crosses the limit. Multiple bounded lines can share a chunk.
+
 - Envelope: **JSON-RPC 2.0** over newline-delimited messages.
 - Protocol version constant: **`1.9.0`**
   (`AGENC_DAEMON_PROTOCOL_VERSION` in `runtime/src/app-server/protocol/index.ts`).
--- a/runtime/src/app-server/transport/stdio.ts
+++ b/runtime/src/app-server/transport/stdio.ts
@@ -12,7 +12,6 @@
  */
 
 import { Buffer } from "node:buffer";
-import { createInterface, type Interface } from "node:readline";
 import type { Readable, Writable } from "node:stream";
 import type { JsonObject, JsonValue } from "../protocol/index.js";
 import {
@@ -21,15 +20,8 @@ import {
   maxQueuedRequestsFromOptions,
 } from "../overload.js";
 import { isRecord } from "../../utils/record.js";
+import { BoundedJsonLineReader } from "../../utils/bounded-json-lines.js";
 
-/**
- * Default upper bound on a single unterminated input line, matching the
- * websocket transport's default max payload. A peer that streams bytes
- * without ever emitting a newline would otherwise grow the readline
- * internal buffer (and daemon memory) unbounded; the transport tracks the
- * bytes seen since the last newline and tears the connection down once this
- * cap is exceeded, treating it as a fatal framing violation.
- */
 export const AGENC_STDIO_DEFAULT_MAX_LINE_BYTES = 16 * 1024 * 1024;
 
 export interface AgenCStdioTransportOptions {
@@ -55,7 +47,7 @@ export class AgenCStdioTransport {
   // request that authenticates and establishes connection state.
   #initializeBarrier: Promise<void> = Promise.resolve();
   #queuedNormalMessages = 0;
-  #reader: Interface | null = null;
+  #reader: BoundedJsonLineReader | null = null;
 
   constructor(options: AgenCStdioTransportOptions) {
     this.#options = options;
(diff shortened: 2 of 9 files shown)


7etsuo:
Validation for 376716c9558e28d00ec6351f7ea4e8ae6aaf78df is complete. Both TypeScript gates, 103 focused tests, 92 Linux native checks, build, real-PTY startup, and the full local runtime suite of 23,941 tests pass.

Cursor Bugbot completed its independent review with no issues after 8 minutes 39 seconds. The approval router had already exceeded its 8-minute wait and left a COMMENTED review, not an approval. The completed Bugbot review satisfies the independent-review requirement. Security review succeeded, and Sonar reports an OK gate with zero new issues or duplication.

No hosted test workflow was enabled or run. No review or check was rerun. Main has no required status or review gate, and this is a manual exact-head squash merge without an admin override.

PR #2339, opened 2026-09-09
fix(thread-store): publish live writers after registry persistence

### LINKED ISSUE #2088: fix(thread-store): roll back live writer registration when persistence fails
## Summary

`FileThreadStore` publishes a live recorder before its registry update is durable. If registry locking, reading, or SQLite upsert fails, the method throws but leaves a live writer registered for a thread that the durable index does not contain.

## Current behavior

[`createThread`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/thread-store/store.ts#L417-L473) inserts the rollout recorder into `liveRecorders` at line 429, then calls `updateRegistry`.

[`resumeThread`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/thread-store/store.ts#L476-L533) inserts the recorder inside the registry mutator at line 495, before persistence completes.

[`updateRegistry`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/thread-store/store.ts#L1245-L1252) invokes the mutator and then writes registry entries. [`writeRegistryUnlocked`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/thread-store/store.ts#L1285-L1289) performs SQLite upserts. Neither create nor resume rolls back `liveRecorders` when any later step throws.

Open PR #2041 changes the helper used to bind the recorder but still binds it before registry persistence and adds no rollback.

## Reproduction

1. Create a thread store with a valid rollout recorder.
2. Inject a failure in registry persistence after the registry mutator runs.
3. Call `createThread` and observe the persistence error.
4. Retry `createThread` with the same ID.

The retry fails with `thread <id> already has a live local writer`, although no registry row was committed. The same ordering affects `resumeThread`.

## Expected behavior

A failed create or resume should restore the exact pre-call in-memory state. Once the method throws, a retry must be able to reach persistence again.

## Impact

One transient registry failure can leave the thread unusable for the lifetime of the store. Live-writer operations and registry reads can also disagree about whether the thread exists.

## Suggested fix

- Persist the registry entry before publishing the recorder, or wrap publication in a `try`/`catch` that restores the previous map entry exactly.
- Treat replacement of an existing binding separately from insertion so rollback does not close or lose a prior valid writer.
- Apply the same transaction order to create and resume.

## Acceptance criteria

- Injected lock, registry-read, and SQLite-upsert failures leave no new live recorder.
- A failed create or resume can be retried successfully with the same ID.
- An existing valid recorder remains unchanged if a later operation fails.
- Registry reads and live-writer lookup agree after both success and failure.
- Tests exercise failures before the mutator, after the mutator, and during upsert.

## Duplicate check

Searched open and closed issues and pull requests for thread-store live writers, failed registry persistence, rollback, and phantom thread registration. No matching report was found. PR #2041 does not change the failure ordering.

7etsuo:
## Summary

- Prepare the existing rollout index before installing a live writer. Publish the writer and its flush callback only after the registry update returns successfully.
- Apply the same order to create and resume, preserving their existing archive checks and metadata/index ordering.
- Leave existing callbacks and live writers unchanged when indexing, registry reads, locking, or SQLite upsert fails. The same thread ID can be retried after the fault is repaired.

## Validation

- Original source reproduces four failures with four passing controls across create and resume. The regressions check the public append path, not private map contents.
- Both runtime typechecks and 216 tests across seven files pass, including backfill, multi-project routing, daemon contracts, and bootstrap services. These include direct rollout-flush mirror coverage and callback ownership after failed indexing/registry reads.
- On 8e349b397ed2bc210ff435eb307ee56c36e54ea9, dedicated new-test typecheck, build, package entrypoints, generated SDK types, and real-PTY startup pass. The full local suite passes 24,212 runtime tests across 2,306 files, plus root and launcher checks. Linux native checks pass 92 tests across five files.
- Exact-head Cursor Bugbot completes without findings, Cursor explicitly approves that SHA, and Security succeeds. Sonar is OK with zero new issues, zero security hotspots, and zero duplication.
- GitNexus rates FileThreadStore HIGH with 16 direct callers and two indexed bootstrap flows; interface dispatch means this is a lower bound. The implementation changes only create/resume publication order and splits index preparation from callback registration. No measured FND input changed.

This change does not make the existing whole-registry upsert loop transactional or undo index rows already committed from a valid rollout. It prevents a failed registration from publishing a new in-memory writer or replacing the prior callback. No schema or public API changes are needed.

Hosted Actions remain disabled. No Actions dispatch, retry, or rerun is requested.

Closes #2088.

### DIFF
--- a/runtime/src/thread-store/store.ts
+++ b/runtime/src/thread-store/store.ts
@@ -426,7 +426,7 @@ export class FileThreadStore implements ThreadStore {
       params.source === undefined
         ? undefined
         : canonicalizeThreadSource(params.source);
-    this.bindLiveRecorder(threadId, params.rolloutStore);
+    this.prepareLiveRecorder(params.rolloutStore);
 
     this.updateRegistry((registry) => {
       const now = new Date().toISOString();
@@ -471,6 +471,7 @@ export class FileThreadStore implements ThreadStore {
       };
       registry.set(threadId, entry);
     });
+    this.bindLiveRecorder(threadId, params.rolloutStore);
   }
 
   resumeThread(params: ResumeThreadParams): void {
@@ -492,7 +493,7 @@ export class FileThreadStore implements ThreadStore {
         );
       }
 
-      this.bindLiveRecorder(threadId, params.rolloutStore);
+      this.prepareLiveRecorder(params.rolloutStore);
 
       const now = new Date().toISOString();
       const entry: RegistryEntry = {
@@ -531,6 +532,7 @@ export class FileThreadStore implements ThreadStore {
       };
       registry.set(threadId, entry);
     });
+    this.bindLiveRecorder(threadId, params.rolloutStore);
   }
 
   appendItems(params: AppendThreadItemsParams): void {
@@ -1174,12 +1176,15 @@ export class FileThreadStore implements ThreadStore {
    * (`appendItems`, metadata patches, path lookups).
    */
   private bindLiveRecorder(threadId: ThreadId, rolloutStore: RolloutStore): void {
-    this.liveRecorders.set(threadId, rolloutStore);
     rolloutStore.setOnRolloutCommitted((rolloutPath) => {
       if (this.closed) return;
       if (this.liveRecorders.get(threadId) !== rolloutStore) return;
       this.indexRolloutFile(rolloutPath);
     });
+    this.liveRecorders.set(threadId, rolloutStore);
+  }
+
+  private prepareLiveRecorder(rolloutStore: RolloutStore): void {
     if (existsSync(rolloutStore.rolloutPath)) {
       this.indexRolloutFile(rolloutStore.rolloutPath);
     }
--- a/runtime/tests/thread-store/writer-registration-retry.test.ts
+++ b/runtime/tests/thread-store/writer-registration-retry.test.ts
@@ -1,0 +1,133 @@
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, describe, expect, test, vi } from "vitest";
+import { RolloutStore } from "../../src/session/rollout-store.js";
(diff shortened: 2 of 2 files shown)


7etsuo:
Final validation for 8e349b397ed2bc210ff435eb307ee56c36e54ea9

- Original source reproduces four failures with four passing controls. The new regressions cover create/resume registry read, lock, upsert, and initial-index failures, retrying the same ID, callback preservation, and rejection of a second live writer.
- Both runtime typecheck configurations pass. All 216 targeted tests across seven files pass, including existing direct-flush mirror, backfill, multi-project, daemon, and bootstrap tests. The new test file typechecks independently without suppressions.
- Build, package entrypoints, generated SDK types, and real-PTY startup pass. The full local suite passes 24,212 runtime tests across 2,306 files, plus root and launcher checks. Linux native checks pass 92 tests across five files. All checks ran on the exact head with a clean worktree.
- Exact-head Bugbot completes without findings, Cursor explicitly approves that SHA, and Security succeeds. Sonar is OK with zero new issues, zero security hotspots, and zero duplication.
- GitNexus class impact is HIGH and was reported before editing. The staged change is confined to two files and the expected creation, resume, binding, preparation, and test symbols. FND artifact SHA256 remains 09619573ef5cb56c9a6dbd1818430956a323e95a687b4cd3efdda988d5073c00; none of its measured inputs changed.
- This PR head has zero GitHub Actions runs. Automatic test workflows remain disabled, and release/bootstrap workflows remain operator-only. No Actions dispatch, retry, rerun, auto-merge, admin override, or hook bypass was used.

The existing whole-registry upsert loop remains nontransactional. The fix defers in-memory writer and callback publication; it does not undo index rows already committed from a valid rollout.

Local logs are under /tmp/user/1000/agenc-issue-sweep-20260907/2088-*.log.

PR #2352, opened 2026-09-09
fix(permissions): invalidate approvals across session resets

### LINKED ISSUE #2127: fix(permissions): invalidate in-flight network approvals on session clear
## Summary

`clearSessionHosts()` empties the network decision caches but leaves requests already awaiting review alive. If one of those requests later receives a session-level decision, its owner writes the old decision back into the newly cleared cache.

## Verified current behavior

- [`clearSessionHosts`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/permissions/network-approval.ts#L394-L404) clears only the allow and deny sets and explicitly leaves pending approvals untouched.
- The pending owner awaits the resolver, then unconditionally updates those sets in [`requestNetworkApproval`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/permissions/network-approval.ts#L522-L558).
- `/clear` calls `clearSessionHosts` in [`commands/clear.ts`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/commands/clear.ts#L132-L135).

The reproduction starts a request with a deferred resolver, waits until it owns the pending entry, calls `clearSessionHosts()`, and then resolves `approved_for_session`. The first request returns allow, `sessionAllowedSize` becomes 1 after the clear, and a second request with no resolver is allowed from the repopulated cache.

## Impact

A decision initiated for the previous chat state can survive `/clear` and govern later requests. The reset does not provide the session boundary promised by the command.

## Fix direction

Give the service a session generation. Pending owners must capture the generation and refuse to cache or publish session-scoped results after it changes. Clearing should also settle or cancel old waiters so stale work does not remain attached to the new session.

## Acceptance criteria

- A resolver result arriving after `clearSessionHosts()` cannot populate either session cache.
- Requests begun after the clear do not join pending work from the previous generation.
- Old waiters terminate with a defined cancelled or denied result.
- Tests cover late allow, late deny, concurrent waiters, and repeated clears.

## Duplicate check

GitHub issue and pull-request searches for `clearSessionHosts`, pending network decisions, session reset races, and `/clear` found no matching report.

MarcoDotIO:
Closes #2127.

`/clear` must cancel approvals from the previous conversation. Give each session its own pending map and cancellation signal so late resolver decisions cannot repopulate caches or remove newer requests. Owners and waiters return `session_cleared`, including when reset arrives during asynchronous owner cleanup. Preserve safe handling of native/frozen abort reasons and promises that reject after cancellation.

Rebased onto `main` at `95d3c748a`, including merged #2358 and #2359. The remaining hook and guidance changes complement those fixes:

- Keep upstream's canonical hook grant implementation and serializer coverage; replace the overlapping synthetic permission probe with real `CanonicalFileReadTool` integration tests for main/subagent/escaped transcripts, unrelated-file denial, and unchanged parent permissions.
- Preserve the `exec_command`, `Edit`, and `FileRead` examples from upstream, generate them through the shared serializer, and validate printed guidance with both parsers and config validation. Correct remaining shell-prefix and empty-parentheses suggestions.
- Separate policy/cache checks from pending approval coordination to make the cancellation code easier to review.

Validation on macOS arm64, Node 26.5.0, npm 11.17.0:

- All 151 targeted tests passed across network approvals, `/clear`, hooks, guidance, and tool-name authority. The new cleanup-window regression failed before its fix and passes afterward.
- Runtime and test-support typechecks passed.
- Production build, package entrypoint checks, and SDK generated-type checks passed with Xcode clang and its matching macOS 26.5 SDK.
- Full focused permission/settings/hook/clear suite: 1,243 passed, 3 failed. The same three failures reproduce on clean `main` at `95d3c748a`: the donor import-boundary assertion for `permissions/read-only-grant.ts`, the audit-log `/tmp` versus `/private/tmp` assertion, and the special temporary-directory permission grant assertion.
- `npm run test:fast -- --base origin/main` stops in the ripgrep preflight with `Cannot access 'AGENC_LINUX_SANDBOX_ARG0' before initialization`; the identical preflight failure reproduces on clean `main` at `95d3c748a`.

Native Windows execution and the full Linux release boundary were not run locally. Windows absolute-path serialization and the portable real-file hook integration cases remain covered by the targeted tests.


7etsuo (commented):
Rebase after #2358 and #2359, resolve `runtime/src/permissions/network-approval.ts`, and reconcile the overlapping hook and guidance changes. Keep the real `CanonicalFileReadTool` tests. Rerun the session-reset cancellation and late-approval tests, then request review of the resolved permission code.

MarcoDotIO:
Let me know if there are any other changes needed.

PR #2363, opened 2026-09-10
fix(permissions): recognize signed session plan files in checkPermissions

### LINKED ISSUE #2131: fix(permissions): recognize signed internal files during permission checks
## Summary

The system file tools do not recognize signed active-session files during `checkPermissions`. They ask for approval when the target is outside the workspace even though the execution boundary already verifies and permits the same target.

## Verified current behavior

At commit `6d9243b2820821ac0443eaabee58fc89d156b688`, a hermetic reproduction created an active plan file under a separate `AGENC_HOME`, then passed its signed session ID and home arguments to a `Write` tool configured for a different workspace root.

With permission mode `acceptEdits`:

- `Write.checkPermissions` returned `behavior: "ask"` with a `workingDir` decision.
- Calling the tool's execution boundary with the same signed input succeeded.

The mismatch comes from two different path policies:

- [`Write.checkPermissions`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/tools/system/file-write.ts#L333-L355), [`Edit.checkPermissions`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/tools/system/file-edit.ts#L856-L878), and [`FileRead.checkPermissions`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/tools/system/file-read.ts#L1511-L1535) all call the common path validator without checking the signed active-session context.
- The established permission pipeline explicitly checks internal editable paths before dangerous-path and workspace gating in [`utils/permissions/filesystem.ts`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/utils/permissions/filesystem.ts#L1291-L1304).
- The execution boundary verifies the signed session context and permits only the active plan file in [`safePathAllowingSessionPlanFile`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/tools/system/filesystem.ts#L1167-L1189).

Existing execution tests cover direct reads, writes, and edits of an active plan outside the workspace, but they do not run each tool's permission check first.

## Impact

Normal plan maintenance can surface an unnecessary approval even in `acceptEdits`. In a daemon, SDK, or unattended flow without an approval responder, the permission phase can block an operation that the signed execution policy considers valid. The behavior also differs depending on whether a filesystem operation uses the established or system-tool permission path.

## Fix direction

Teach the common system-tool permission path to validate the signed active-session context and apply the same narrow internal-file exceptions as the execution boundary. Keep this exception after explicit deny rules and limit it to the exact verified active-session path. Do not turn `AGENC_HOME` or its parent directory into a general allowed root.

This is related to #2129 but tracks the opposite side of the policy drift: #2129 covers sensitive paths that are allowed too broadly, while this issue covers signed internal paths that are rejected too broadly.

## Acceptance criteria

- `FileRead`, `Write`, `Edit`, `MultiEdit`, and apply-patch permission checks recognize their supported signed active-session paths without asking solely because those paths are outside the workspace.
- Explicit deny rules still win over every internal-path exception.
- Invalid signatures, another session's plan, and unrelated files under `AGENC_HOME` still require approval or are denied as appropriate.
- Permission-plus-execution tests cover active and inactive plan paths outside the workspace for each affected file-tool family.
- The internal-path decision is shared with, or demonstrably equivalent to, the execution-boundary check so the two policies cannot drift independently.

## Duplicate check

GitHub issue and pull-request searches for plan-file permission prompts, signed internal files, and active session plan paths outside the workspace found no matching report. Issue #2129 is related but covers missing protected-path denials rather than missing internal-path allows.

turkonthelurk:
## Changes

Closes #2131.

`FileRead` / `Write` / `Edit` / `MultiEdit` / `apply_patch` `checkPermissions` asked for approval when the target was the signed active-session plan file outside the workspace. Execute already allowed that path via `safePathAllowingSessionPlanFile`.

- After explicit deny rules, `checkToolPathPermission` calls `matchesVerifiedSessionPlanFile`, which reuses the execution-boundary verifier (session HMAC + injected `AGENC_HOME`).
- Invalid signatures, another session's plan, and sibling files under `AGENC_HOME` still ask. Deny rules still win. `AGENC_HOME` is not a general allowed root.

Does not change the plan-mode mutating-tool gate (open Cursor PR #2254). That is a different layer.

## Verification

- `runtime/tests/permissions/signed-session-plan-file.test.ts` plus `path-validation.test.ts` 28/28 (hermetic vitest, Node 26.7.0).
- `npm run typecheck` on the runtime workspace.

### DIFF
--- a/runtime/src/permissions/path-validation.ts
+++ b/runtime/src/permissions/path-validation.ts
@@ -33,6 +33,7 @@ import {
 import {
   withSignedAllowedRoots,
 } from "../agents/_deps/filesystem-args.js";
+import { matchesVerifiedSessionPlanFile } from "../tools/system/filesystem.js";
 import type {
   PermissionDecisionReason,
   PermissionResult,
@@ -666,6 +667,17 @@ export function checkToolPathPermission(
       behavior: "deny",
       message: `Permission to ${verb} ${opts.path} has been denied.`,
       decisionReason,
+    };
+  }
+
+  if (matchesVerifiedSessionPlanFile(opts.path, opts.cwd, opts.input)) {
+    return {
+      behavior: "allow",
+      updatedInput: opts.input,
+      decisionReason: {
+        type: "other",
+        reason: "Signed active-session plan file",
+      },
     };
   }
 
--- a/runtime/src/tools/system/filesystem.ts
+++ b/runtime/src/tools/system/filesystem.ts
@@ -43,7 +43,7 @@ import {
   rmSync,
   writeFileSync,
 } from "node:fs";
-import { resolve, dirname, basename, join } from "node:path";
+import { resolve, dirname, basename, join, isAbsolute, normalize } from "node:path";
 import { resolveHomeContext } from "../../config/home.js";
 // Imported from the defining module rather than the `memory/index.js` barrel.
 // The barrel re-exports the recall pipeline, which reaches `utils/ide.ts` and
@@ -1204,6 +1204,23 @@ export function verifiedPlanFileContextFromArgs(
     ...ctx,
     agencHome: resolveHomeContext({ AGENC_HOME: injectedAgencHome }).path,
   };
+}
+
+/**
+ * True when `targetPath` is the signed active-session plan-file family.
+ * Same authority as {@link safePathAllowingSessionPlanFile}: verified
+ * session id + injected home, exact plan prefix, no AGENC_HOME widening.
+ */
+export function matchesVerifiedSessionPlanFile(
+  targetPath: string,
+  cwd: string,
+  args: Record<string, unknown> | undefined,
+): boolean {
+  const planCtx = verifiedPlanFileContextFromArgs(args);
+  if (planCtx === null || hasUnsafeShape(targetPath)) return false;
+  const absolute = isAbsolute(targetPath) ? targetPath : resolve(cwd, targetPath);
+  const identity = normalizeFilesystemUnicodeIdentity(normalize(absolute));
+  return isSessionPlanFile(identity, planCtx);
 }
 
 export async function safePathAllowingSessionPlanFile(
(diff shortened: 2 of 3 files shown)


7etsuo (changes requested):
The signed-plan shortcut in [path-validation.ts:673](https://github.com/tetsuo-ai/agenc-core/blob/8bac04a8990a9a5bc6679ea50185a113106e7586/runtime/src/permissions/path-validation.ts#L673) preserves explicit `deny` but overrides a matching `ask` rule. A signed active-plan read outside the workspace is allowed even when its exact path requires approval. Preserve `ask` and `deny` before granting the working-directory exception, using the current main session-plan authority. Add a regression with a path-specific `FileRead` ask rule for the active plan and verify that approval is still required.

7etsuo:
Closing as superseded. Thanks for pinning this down: the permission layer really did ask for approval on the signed active-session plan file while `execute` accepted it, and that gap is now closed on `main`.

Two changes covered it through a different mechanism:
- #2365 added a session-derived authority (`runtime/src/planning/session-plan-authority.ts`, consumed by `isPathAllowed`) and wired it into Write, Edit and MultiEdit.
- #2522 wires the same authority into FileRead and apply_patch, the two tools this PR also covered, with parity, negative, rule-precedence and end-to-end tests in `tests/permissions/session-plan-mutation.test.ts`.

The session-derived authority was preferred over deriving it from the tool arguments: the router strips model-supplied `__agenc*` keys before dispatch and the runtime injects the signed ones only at execute time, so at permission time those arguments are not present, and the session object gives the same identity without depending on them. #2131 is closed by #2522.

PR #2374, opened 2026-09-10
fix(runtime): repair planning authority and interactive cancellation

7etsuo:
## Changes

- Make Enter honor the selected approval, retain denial feedback, and resolve approvals by canonical conversation owner.
- Keep cancelled and denied turns stopped across child receipts, queued work, restart, and durable recovery. Continue interrupting descendants if recording the stop fails.
- Allow plan-mode delegation with inherited read-only authority, trusted native tools, guarded reads, and mandatory shell isolation. Switching the parent to YOLO does not widen an existing planning worker.
- Restore cold-start workflow controls and daemon-backed rewind previews.
- Correct Markdown sizing and retry identities, and reject deterministic shell-policy failures before asking for approval.
- Add regression tests, a real-kernel constrained-worker test, and a PTY Escape test with foreground and descendant processes.

## Verification

- Post-rebase runtime run passed 25,727 of 25,730 tests. The clean-worktree fuzzy benchmark passes. Remaining failures are two pre-existing inventory checks and stale FND evidence.
- Runtime and SDK builds, typechecks, generated SDK parity, agent-surface checks, startup in normal and combined YOLO mode, real PTY Escape cancellation, and Linux read-only integration passed.
- Independent review completed. Three later review concerns were confirmed and corrected before merge: stop release now happens only at durable human admission; legacy stops also fence unrelated or later automatic orphan turns; constrained workers retain role instructions without gaining authority.
- Revert-sensitive tests cover all three corrections. The daemon/recovery gate passes 431 tests and the role/delegation gate passes 112.
- Final full runtime verification passed 25,761 of 25,764 tests across 2,414 files, with no skips. Only the two known inventory checks and stale FND evidence fail. Final build, default/combined-YOLO startup, and foreground/descendant Escape cancellation also pass.
- A fourth security finding is fixed: constrained Git object reads now require unrestricted original filesystem read authority. A real native regression reproduced the prior disclosure; the corrected native test and 232 related tests pass. Independent review found no remaining blocker.
- FND evidence capture requires a source revision already on the default branch. A follow-up evidence refresh will use the merged source, without editing recorded hashes.
- Existing launcher configuration and AppArmor profile-name assertions remain documented failures. Native kernel observations confirmed filesystem and network restrictions; no host security rule was weakened.
- macOS and Windows native execution and the fresh live Grok session are not yet verified. Windows constrained shell inspection fails closed; native readers remain available.

Hosted CI stays disabled. No release workflow is dispatched.

### DIFF
--- a/docs/reference/agents.md
+++ b/docs/reference/agents.md
@@ -51,7 +51,8 @@ and searches remain available within the worker's read permissions.
 
 Constrained workers reject `git status` and `git diff`, which can invoke
 repository-configured conversion filters. Object-only Git inspection is
-available when no read-path denial applies; Git objects can otherwise reveal
+available only when the original filesystem authority permits unrestricted
+reads and no read-path denial applies. Git objects can otherwise reveal
 denied file contents. PDF reads that require an external converter are also
 unavailable to these workers. These restrictions do not change ordinary
 coding or verification workers.
--- a/docs/reference/agents.md
+++ b/docs/reference/agents.md
@@ -33,6 +33,34 @@ in `runtime/src/bin/model-facing-tools.ts`, not inside `v2/index.ts`.
 | `show_csv_job_review` | Read one bounded review record |
 | `resolve_csv_job_review` | Approval-gated operator resolution with canonical evidence |
 
+### Read-only planning workers
+
+Plan mode can spawn workers with `isolation: "none"`. AgenC assigns those
+workers a permanent read-only constraint, regardless of the requested role.
+The built-in `Plan` and `scanner` roles have that constraint in every mode.
+Their descendants inherit it, including after reconnect or restoration.
+
+Constrained workers can read and search files, run supported literal
+inspection commands, and coordinate their own constrained descendants. They
+cannot edit files, create worktrees, schedule work, request wider permissions,
+or direct writable workers. Shell inspection uses direct arguments, a
+sanitized environment, and required read-only process isolation with network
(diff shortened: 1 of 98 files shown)


7etsuo:
All four concrete review findings are corrected before merge.

- Daemon stop release now uses a private generation token and happens only
  after canonical human input is durably admitted. Failed submission and
  intervening stops remain held. Seven regressions failed before the fix;
  the expanded gate passes 431 tests.
- Recovery now carries inferred legacy stop state across unrelated orphan
  turns and records checkpoint-time holds. Ten new cases failed before the
  fix; the focused and broader recovery gates pass.
- Planning and custom role instructions remain present alongside read-only
  restrictions, inside the existing untrusted-guidance wrapper. All four
  role-prompt cases failed before the fix and pass after it.
- Git object inspection now requires genuine unrestricted original read
  authority, in addition to the tool-rule gate. Path/glob denials, implicit
  allowlists, and opaque external policies fail closed. A native test
  reproduced the prior committed-file disclosure and confirms the fix.
  The related gate passes 232 tests and the native test passes without skips.

Independent review found no remaining blocker in these corrections. The
final full runtime suite is running before manual merge.

The two Sonar string-sort findings do not indicate a defect here. Both sorts
operate on strings and intentionally use deterministic code-unit ordering
for search pagination and canonical operation identity, not locale-specific
display collation. No suppression or behavior change was added.

PR #2391, opened 2026-09-11
fix(llm): warn on quiet streams and abort only dead sockets

### LINKED ISSUE #2200: A provider stream that goes quiet holds the turn indefinitely and the app cannot tell a stall from long reasoning
## What happens

Desktop soak on xAI grok-4.6 at `reasoning_effort = xhigh`, 2026-09-06. Several turns sat in a sampling request that produced almost nothing for a long time:

| run | turn | what the app showed | what happened |
| --- | --- | --- | --- |
| plan-3, prompt 1 | `sub-conv-mtpaj1h5-2` | `thinking · 9m 51s · 26 tokens` | the phase ended and the model went on to 300 read-only calls |
| plan-3, prompt 1 | same | `considering · 22m 18s · 18 tokens` | the prompt had crossed the context limit; the request was never served (#2196 for the compaction half) |
| plan-3, prompt 2 | `sub-conv-mtpaj1h5-3211` | `working it out · 25m 37s · 30 tokens` | stopped by the driver's cap |
| swarm-2, prompt 1 | `sub-conv-mtpcsux5-2` | `mulling it over · 19m 58s · 26 tokens` | then 27 agents started working |

`resolveStreamIdleTimeoutMs` returns 0 by design (`stream-watchdog.ts`: "a positive operator value is required to install a deadline"), so a stream that stops producing tokens holds the turn until the user presses stop. The ticker shows an elapsed time and a token count; nothing separates a stalled stream from reasoning that has not surfaced a token yet, and two of the four cases above were legitimate.

## Why it matters

The user's only decision is whether to press stop, and the app gives no basis for it. A hard default deadline is the wrong fix: high-effort reasoning on this provider legitimately goes quiet for ten minutes and a deadline would cut real work.

## Proposal

1. A soft signal, not a deadline: after N minutes without a delta (N configurable, default in the range of the legitimate quiet phases seen here, about 10 minutes), the runtime emits a warning event ("no output from the model for N minutes") and the desktop shows it on the ticker with the stop button, so the user decides with information.
2. Provider-level liveness: record whether the HTTP stream is receiving keepalive frames or nothing at all; a dead socket (no bytes of any kind for N minutes) is a stall and can be retried through the existing reconnect ladder without cutting reasoning that is sending heartbeats.
3. Surface `reasoning_effort` on the ticker for these turns, since it is the variable that predicts the quiet phase.

Soak finding F52 in `docs/eval/app-soak-2026-09-05.md`; related F53 (a plan prompt that never presented a plan in 30 minutes) and F55/#2196.

FalconOrtiz:
## Why

A provider stream that goes quiet holds the turn, and the app cannot tell a stall from long high-effort reasoning. `resolveStreamIdleTimeoutMs` returns 0, so `installStreamWatchdog` was a no-op. A positive `stream_watchdog_timeout_ms` aborted any silence, which would cut grok-4.6 xhigh phases that already produced tokens. The ticker hid its stall note while the status said thinking. Soak finding F52 / #2200.

## Scope

`classifyStreamLiveness` in `runtime/src/llm/stream-watchdog.ts` splits `live`, `quiet` (had bytes, no model delta), and `dead` (no bytes at all).

`kick("bytes")` is a heartbeat. It keeps a dead-socket abort from firing and does not reset the quiet warning. `kick("delta")` resets the warning window. `stream-model` classifies each `LLMStreamChunk`. The HTTP client kicks bytes on every body chunk, including empty keepalives.

A zero abort timeout still warns when `onWarning` is set. `stream-model` emits `warning` with cause `stream_idle_warning`. The TUI ticker shows that copy even while thinking, with the existing stop affordance. Transcript allow-lists the cause.

Out of scope: ripgrep, neovim lifecycle, `turn_failed`, reconnect ladder changes.

## Tradeoffs

A configured idle timeout no longer aborts after the first byte. That is the point. Dead sockets that never receive bytes still abort as retryable `stream_idle`. Quiet reasoning after tokens is a user decision (stop), not a hard kill.

When abort is armed, the warning still fires at half the timeout so existing half-time tests stay true. When abort is disabled, the warning uses `STREAM_QUIET_WARNING_MS` (ten minutes).

## Blast Radius

Session streaming (`stream-model`), provider HTTP streams (`client-session`), and the TUI spinner ticker. Review delegates that opt out of the ambient abort still get the soft warning. No change to turn failure terminals.

## Verification

Reproduced on `installStreamWatchdog({ timeoutMs: 0 })`. `timeoutMs` is 0 and neither warning nor abort fires after 500ms. After the fix, `timeoutMs: 0` with `warningMs: 100` warns and does not abort. A dead socket with timeout 100 still aborts. Heartbeats keep a quiet stream alive.

Focused hermetic vitest, Node v26.7.0 from `.tools\node-v26.7.0`:

- `tests/llm/stream-watchdog.test.ts` 6 passed
- `tests/llm/stream-watchdog.silent-generation.test.ts` 7 passed
- `tests/llm/api/watchdog.test.ts` 1 passed
- `tests/phases/stream-model.test.ts` 38 passed, including quiet-vs-dead
- `tests/tui/components/spinner/SpinnerAnimationRow.liveness.test.tsx` 14 passed
- `tests/llm/client-session.test.ts` 32 passed
- `tests/llm/providers/grok/adapter.stream-liveness.test.ts` 3 passed
- `tests/llm/providers/grok/stream-cancellation.test.ts` 8 passed

Closes #2200

### DIFF
--- a/runtime/src/llm/client-session.ts
+++ b/runtime/src/llm/client-session.ts
@@ -1099,7 +1099,7 @@ export class ProviderHttpClientSession {
               if (next.done) return;
               // LLM-09: empty chunks still count as body progress for idle
               // watchdog (providers may send keepalives).
-              watchdog.kick();
+              watchdog.kick("bytes");
               if (!next.value || next.value.length === 0) continue;
               if (prepared.continuation) {
                 const decoded = decoder.decode(next.value, { stream: true });
--- a/runtime/src/llm/stream-watchdog.ts
+++ b/runtime/src/llm/stream-watchdog.ts
@@ -5,25 +5,13 @@
  * `streamWatchdogFiredAt`, `streamIdleAborted`) driven by
  * `stream_idle_timeout_ms` from provider info.
  *
- * The watchdog has no implicit deadline. The canonical config snapshot carries
- * a positive timeout when operators opt in; `0` disables it. This keeps long
- * silent reasoning and tool-argument generation valid for arbitrarily long
- * turns.
+ * Abort is opt-in (`timeoutMs` > 0) and fires only when the socket never
+ * received bytes. Heartbeats (`kick("bytes")`) keep that abort from firing.
+ * Model deltas (`kick("delta")`) reset the quiet-warning window. A zero abort
+ * timeout still warns when `onWarning` is set, so long grok-4.6 xhigh quiet
+ * phases stay alive with a ticker signal instead of a hard kill.
  *
- * Timers use monotonic clock (I-82) via `monotonicMs()` — immune to
- * NTP corrections, `date` set, suspend/resume, container clock skew.
- *
- * ## Usage patterns
- *
- * **Streaming** (T7 `chatStream`): call `kick()` on every received
- * chunk. The watchdog fires abort after `STREAM_IDLE_TIMEOUT_MS` of
- * silence since the last kick.
- *
- * **Total-timeout fallback** (T5 `chat()`): install with no kicks —
- * the watchdog fires after `STREAM_IDLE_TIMEOUT_MS` from install.
- * This is a coarse fallback because we can't observe intra-response
- * progress without a streaming channel; T7 replaces it with real
- * per-chunk kicks once `chatStream` wires in.
+ * Timers use monotonic clock (I-82) via `monotonicMs()`.
  *
  * @module
  */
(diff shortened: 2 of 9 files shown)


7etsuo (changes requested):
In [stream-watchdog.ts:250](https://github.com/tetsuo-ai/agenc-core/blob/7cb97a5b0610a8dddcc6263b4fd5925749f9ba46/runtime/src/llm/stream-watchdog.ts#L250), the first byte permanently disables the configured idle abort. If a provider sends one chunk and then leaves the connection open, a run without a separate provider deadline can hang indefinitely. Measure the abort interval from the most recent byte and keep the model-silence warning separate. Add a regression that sends one chunk, stalls past the configured timeout, and asserts cancellation.

turkonthelurk:
Review (comment only, not an approval).

Quiet-vs-dead split matches what #2200 still needed (soft warning, not a hard abort after the first byte). Base is behind current `main`; rebase before land. No merge from this comment.

FalconOrtiz:
﻿@7etsuo addressed CHANGES_REQUESTED on stream-watchdog:

- Abort interval now measures from the most recent byte (not permanently disabled after the first).
- Quiet/model-silence warning stays on the delta clock.
- Regression: one chunk, stall past timeout, asserts cancellation.

Head: a88faeb0dd1480e54fcd89b5925a863b5d2de7e5 — please re-review when convenient.

FalconOrtiz:
﻿e2e_smoke_ok on feeaa92544f3ea8c3b456b7b8f192e7f1c0a991c.

Last-byte idle abort is in production. The stream-model suite now warns at half-timeout then aborts from the last byte. Coordinator hermetic 59/59.

@7etsuo please re-review. merge_ready only. No merge from this side.

7etsuo:
- [stream-watchdog.ts:131](https://github.com/tetsuo-ai/agenc-core/blob/feeaa92544f3ea8c3b456b7b8f192e7f1c0a991c/runtime/src/llm/stream-watchdog.ts#L131) reads `chunk.content.length` unguarded, so a chunk without `content` throws in `onChunk`. The `text_delta` stubs in `exec-agent-hook-run-turn.test.ts` send such chunks. Use `(chunk.content?.length ?? 0) > 0` and add a test.
- The runtime warns at `timeoutMs / 2` ([stream-watchdog.ts:110](https://github.com/tetsuo-ai/agenc-core/blob/feeaa92544f3ea8c3b456b7b8f192e7f1c0a991c/runtime/src/llm/stream-watchdog.ts#L110)), 5 minutes by default; the spinner waits 10 ([SpinnerAnimationRow.tsx:305](https://github.com/tetsuo-ai/agenc-core/blob/feeaa92544f3ea8c3b456b7b8f192e7f1c0a991c/runtime/src/tui/components/spinner/SpinnerAnimationRow.tsx#L305)). Use one threshold.
- The abort still fires `timeoutMs` after the last chunk, as on main, but the title, body and [stream-model.ts:15](https://github.com/tetsuo-ai/agenc-core/blob/feeaa92544f3ea8c3b456b7b8f192e7f1c0a991c/runtime/src/phases/stream-model.ts#L15) say quiet streams are not aborted. Retitle to `feat(llm): warn on quiet model streams` and use `Refs #2200`, since provider liveness is not implemented.

FalconOrtiz:
@7etsuo addressed CHANGES_REQUESTED on https://github.com/tetsuo-ai/agenc-core/pull/2391#issuecomment-5809214041:

- `streamChunkHasDelta` now uses `(chunk.content?.length ?? 0) > 0`. Test: `text_delta` stubs without `content` do not throw.
- Quiet warning is one threshold: `timeoutMs / 2` of the default abort (5 min). Spinner shares `STREAM_QUIET_WARNING_MS`.
- Title is `feat(llm): warn on quiet model streams`. Body and `stream-model.ts` I-11 comment state last-byte idle abort remains. `Refs #2200` (provider liveness is not implemented).

Head: 637fe8a70391647a484d3c721c7a635fda6870c8 — please re-review when convenient. No merge from this side.

PR #2400, opened 2026-09-11
fix(session): drop failed compaction payload chunks on open

### LINKED ISSUE #2162: Repair sessions whose compaction payload chunks fail strict validation
## What happens

Before #2160, a compaction whose source history held a secret-like string wrote payload chunks that the rollout line encoder redacted after their byte count and digest were taken. Such a rollout is now permanently unreadable: `readAll` and the strict resume reject the chunk (`canonical compaction_payload_chunk payload does not match the runtime schema`), every turn fails with `payload chunk count does not match its manifest`, and a daemon restart does not help. The attempt itself is already recorded as `compaction_failed` with `reason: commit_failed`, so its payload records carry nothing the session needs.

#2160 stops new occurrences; it does not repair rollouts already written this way.

## Proposal

A repair pass (a `agenc session repair <id>` command, or a step in startup recovery) that, for an attempt with a `compaction_failed` record, drops that attempt's `compaction_payload_chunk` lines (and the persisted intent's payload manifests it can no longer satisfy) through the existing atomic rewrite (`rewriteRolloutExcludingPhysicalLinesAtomically`), leaving the failure record in place. The reader could also treat a chunk whose attempt later failed as inert instead of fatal, but the strict validator runs per line and the fix above keeps its contract.

Example to test against: the soak session `conv-mtnmmso6` in the isolated home (intent at line 8275, two chunks at 8276 and 8277, `compaction_failed` at 8284). Findings F14 in the desktop soak.

FalconOrtiz:
## Why

#2162. Before #2160, a compaction whose source history held a secret-like string wrote `compaction_payload_chunk` rows whose byte count and digest were taken before redaction. Strict resume then rejects the chunk (`canonical compaction_payload_chunk payload does not match the runtime schema`). `readAll` skips the invalid line and then throws `payload chunk count does not match its manifest`. The attempt is already `compaction_failed` with `reason: commit_failed`, so those payload rows are not needed. #2160 stops new writes. This repairs rollouts that already have them.

## Scope

- `SessionStore.rewriteFailedCompactionPayloadChunksAtomically` collects physical `compaction_payload_chunk` rows whose `attempt_id` has a `compaction_failed` record and drops them through `rewriteRolloutExcludingPhysicalLinesAtomically`.
- Exclusion matching uses the physical digest plus JSON `type` / `attempt_id` / `payload_kind`, so a schema-invalid chunk can still be dropped. Kept rows still go through `parseRolloutLine`.
- `RolloutStore.open` runs that rewrite after the store opens and before compaction reconciliation.
- `hydrateManifestCompactionItems` does not reconstruct payload manifests for an attempt that already failed.
- The canonical scanner allows a failed attempt without reconstructed source payloads. `#seal` does the same when `terminal === "failed"`.
- One unique test mutates a `source_history` fragment after `compaction_failed`, asserts the schema error, then resumes and checks that the chunks are gone, the failure record remains, and `keep-me` history reconstructs.

Not changed: MCP SSE, Anthropic wire, plugin loaders, committed-chunk tamper rejection, the strict per-line chunk schema.

## Tradeoffs

The reader still does not treat an invalid chunk as inert. The chunk has to leave the file. That keeps the per-line validator contract and means a failed attempt without chunks is now a valid sealed journal.

## Blast Radius

Resume and startup of sessions that already recorded `compaction_failed` with leftover payload rows. Committed or rollback payload tampering still fails closed, as the existing restart tests require. Physical rewrite stays on the existing atomic path.

## Verification

Hermetic vitest with Node v26.7.0 from `.tools/node-v26.7.0`.

- Before the fix, `repairs failed-attempt payload chunks that fail strict validation` failed on resume (`payload chunk count does not match its manifest`, then `compaction source_history manifest has no chunk chain or durable release`).
- After the fix, `node scripts/run-hermetic-vitest.mjs run tests/session/rollout-store.compaction-transaction.test.ts tests/services/compact/payload-manifest.test.ts` exit 0, 58 passed.
- `tests/session/canonical-rollout-scanner.test.ts`, `tests/state/recovery-journal-contract.test.ts`, `tests/session/rollout-store.canonical-scan-reuse.test.ts` exit 0, 65 passed.
- Descriptor-bound streaming rewrite tests exit 0, 2 passed.
- `rolls back a partial append before degraded requeue` fails on this Windows host against `main` as well (7-byte tail not rolled back). Not introduced here.

Closes #2162

### DIFF
--- a/runtime/src/session/canonical-rollout-scanner.ts
+++ b/runtime/src/session/canonical-rollout-scanner.ts
@@ -241,7 +241,7 @@ interface MutableAttemptScan {
 }
 
 export interface CanonicalCompactionAttemptScan {
-  readonly intent: CompactionIntentV1;
+  readonly intent?: CompactionIntentV1;
   readonly records: readonly StrictCanonicalJournalRecord[];
   readonly admissionValid: boolean;
   readonly hasLaterCanonicalWork: boolean;
@@ -570,7 +570,9 @@ function scanCanonicalRolloutUntimed(
           [...state.attempts].map(([attemptId, attempt]) => [
             attemptId,
             {
-              intent: attempt.intent!,
+              ...(attempt.intent !== undefined
+                ? { intent: attempt.intent }
+                : {}),
               records: Object.freeze(attempt.records.slice()),
               admissionValid: validAdmission(attempt),
               hasLaterCanonicalWork:
@@ -943,7 +945,11 @@ function observeCanonicalRecord(
       attempt.intent === undefined &&
       !attempt.terminal,
   );
-  if (incompleteAttempt !== undefined) {
+  if (
+    incompleteAttempt !== undefined &&
+    (item.type === "compaction_committed" ||
+      item.type === "compaction_rollback_committed")
+  ) {
     throw new Error(
       "canonical compaction intent is missing its required source payload bundle",
     );
@@ -1341,6 +1347,7 @@ function assertAttemptsReconstructed(
   attempts: ReadonlyMap<string, MutableAttemptScan>,
 ): void {
   for (const attempt of attempts.values()) {
+    if (failedAttemptDroppedPayloads(attempt)) continue;
     if (attempt.intent === undefined) {
       throw new Error(
         "canonical compaction intent did not reconstruct its source manifests",
@@ -1360,6 +1367,13 @@ function assertAttemptsReconstructed(
   }
 }
 
+function failedAttemptDroppedPayloads(attempt: MutableAttemptScan): boolean {
+  return (
+    attempt.intent === undefined &&
+    attempt.records.some((record) => record.item.type === "compaction_failed")
+  );
+}
+
 /**
  * Line numbers the digest-anchored second pass must re-read: the caller's
  * additional lines, every attempt's active-history refs, and the live active
@@ -1372,7 +1386,8 @@ function collectSourceLines(
 ): Set<number> {
   const sourceLines = new Set(options.additionalSourceLines ?? []);
   for (const attempt of attempts.values()) {
-    for (const ref of attempt.intent!.source.active_history_refs) {
+    if (attempt.intent === undefined) continue;
+    for (const ref of attempt.intent.source.active_history_refs) {
       sourceLines.add(ref.first_sequence);
     }
   }
--- a/runtime/src/session/rollout-store.ts
+++ b/runtime/src/session/rollout-store.ts
@@ -864,6 +864,9 @@ export class RolloutStore {
       }
 
       this.store.open(meta);
+      this.store.rewriteFailedCompactionPayloadChunksAtomically(
+        COMPACTION_SOURCE_DIGEST_DOMAIN,
+      );
       this.promoteDurableCheckpointSchema(meta);
       this.rebuildLiveToolPairProjection();
       // Re-check under the canonical rollout lease. Retention can retire the
@@ -2563,6 +2566,7 @@ export class RolloutStore {
     );
     for (const attempt of orderedAttempts) {
       const intent = attempt.intent;
+      if (intent === undefined) continue;
       if (this.compactionRetentionRepo.get(intent.attempt_id) !== undefined) {
         continue;
       }
--- a/runtime/src/session/session-store.ts
+++ b/runtime/src/session/session-store.ts
@@ -140,6 +140,131 @@ type RolloutPhysicalLineExclusion = {
     }
 );
 
+function isCompactionPayloadKind(
+  value: string | undefined,
+): value is CompactionPayloadKind {
+  return (
+    value === "active_history_refs" ||
+    value === "source_history" ||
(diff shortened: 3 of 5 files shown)


FalconOrtiz:
awaiting_maintainer

Sonar excellency on `d8516f9827712db526d492f86c5e8b0a0d1b5323`:
- Quality Gate OK
- 0 OPEN new issues
- 0.0% New Code duplication
- 0 security hotspots

e2e smoke receipt: `reports/e2e-smoke-2400-d8516f98.md` (`e2e_smoke_ok`, 58 passed on the failed-compaction payload repair path).

No merge performed.

7etsuo (changes requested):
The cleanup in [rollout-store.ts:867](https://github.com/tetsuo-ai/agenc-core/blob/d8516f9827712db526d492f86c5e8b0a0d1b5323/runtime/src/session/rollout-store.ts#L867) deletes failed payload rows before checking later retention pins. An early failed attempt followed by a committed compaction shifts that compaction's physical source ordinals, breaking source verification and potentially startup recovery. Preserve those ordinals or reject deletions that affect live pins, as the existing pruning guard does. Add a regression covering an early failed attempt, a later pinned compaction, restart, and source release.

turkonthelurk:
Review (comment only, not an approval).

Failed-attempt payload-chunk drop on open looks like the right repair for #2162 (physical digest match so schema-invalid chunks still drop). Rebase vs current `main` before land. No merge from this comment.

FalconOrtiz:
﻿e2e_smoke_ok on 26657ee294cb3a10b783dd5a0571b8a485b5e80c.

Coordinator reviewed the CR diff (live pin ordinals preserved) and re-ran compaction-transaction hermetic 51/51. Smoke matrix 59/59.

@7etsuo please re-review. merge_ready only. No merge from this side.

FalconOrtiz:
﻿@7etsuo re-review requested on `26657ee294cb3a10b783dd5a0571b8a485b5e80c`.

Failed-payload open repair now rejects deletions at or before live retention pin ordinals (same guard shape as physical source pruning). Early failed attempt + later pinned compaction keeps physical source sequences across restart; source release still completes. Regression: `preserves failed-attempt payload ordinals under a later pinned compaction across restart and source release`.

Sonar excellency on this SHA: QG OK, 0 OPEN, 0.0% new duplication. No merge.

7etsuo:
[rollout-store.ts:867](https://github.com/tetsuo-ai/agenc-core/blob/26657ee294cb3a10b783dd5a0571b8a485b5e80c/runtime/src/session/rollout-store.ts#L867) still reads only the current epoch's SQLite pins. After a terminal-run reopen, or with a missing state DB, the drop still shifts live pin ordinals, so open fails after source release. Derive protected ordinals from unreleased canonical intents, skip the drop when no epoch row exists, and add regressions for both cases.

- [canonical-rollout-scanner.ts:948](https://github.com/tetsuo-ai/agenc-core/blob/26657ee294cb3a10b783dd5a0571b8a485b5e80c/runtime/src/session/canonical-rollout-scanner.ts#L948) now accepts any record between an incomplete intent and its chunks. Allow only that attempt's lifecycle rows.
- Before deleting, validate the `compaction_failed` row and check the attempt has no commit or rollback.
- The `Math.max` spread at [session-store.ts:282](https://github.com/tetsuo-ai/agenc-core/blob/26657ee294cb3a10b783dd5a0571b8a485b5e80c/runtime/src/session/session-store.ts#L282) can throw `RangeError`. Use a loop.

FalconOrtiz:
@7etsuo addressed on `cdebd50dd0a569f227c9aa4d7107a202ba40b147`.

- `rollout-store.ts`: skip the failed-payload drop when no lifecycle epoch row exists. Protected ordinals now come from unreleased canonical intents (unioned with any current-epoch SQLite pins). Regressions: terminal-run reopen onto a later SQLite epoch, and missing state DB.
- `canonical-rollout-scanner.ts`: while an intent is incomplete, only that attempt's lifecycle rows are allowed; commit/rollback without payloads still fail closed.
- Before delete: `compaction_failed` must validate, and the attempt must have no commit or rollback.
- `session-store.ts`: live ordinal max uses a loop, not `Math.max` spread.

Focused hermetic: compaction-transaction 56/56, payload-manifest 8/8, session-store ordinal-loop + descriptor-bound rewrite 3/3. No merge.

FalconOrtiz:
Quality Gate 11.3% New Code duplication and typescript:S3776 on collectFailedPayloadRepair are addressed at 4a92377f7.

The cloned failed-payload tests now share one case body. Unique assertions stay.
collectFailedPayloadRepair is split into small helpers. Each helper is at or below cognitive complexity 15. Drop behavior is unchanged.

Focused verify exit 0: node scripts/run-hermetic-vitest.mjs run tests/session/rollout-store.compaction-transaction.test.ts tests/services/compact/payload-manifest.test.ts (64 tests).

PR #2548, opened 2026-09-17
fix(secrets): preserve benign integer arrays in tool history without weakening key redaction

### LINKED ISSUE #2476: [URGENT] fix(secrets): preserve benign integer arrays in durable tool history without weakening credential redaction
## Problem and priority

**Urgent / bug. Assignee: @7etsuo.** The global secret sanitizer identifies any textual bracketed list of **32–200 comma-separated integers, each 0–255**, as wallet key material. Ordinary permutations and byte-valued analytical data therefore become `[REDACTED_SECRET]` in saved tool output and durable conversation history. This silently destroys legitimate evidence needed to inspect, resume, or compact a coding session.

Confirmed in the Terminal-Bench repository build `cec48942cfa97fd095f057779d8c55b6cb39a7f3` and independently reproduced in current main `f037754766c3f0ad543f85a3c3db1eea7afa3b01`. The relevant sanitizer, conversion, serialization, and compaction files are unchanged between those commits. This issue requests a fix; the active benchmark build has not been modified.

## Observed evidence

In job `tb4-grok-cec48942cfa9-clean`, task `interleaved-vigenere`, successful `exec_command` call `call-b481902d-5bcb-4bb4-bf5e-40b36f993d4d-31` computed index permutations from the task's development sample. Saved output includes:

```text
T[:100] [REDACTED_SECRET]
is T a perm of 0..M-1? True
...
invT[:80] (source for each output position of delayed stream) [REDACTED_SECRET]
```

The producing code prints `T[:100]` and `invT[:80]`; these are computed index mappings, not provider credentials or wallet exports. The tool completed successfully with exit code 0. Negative-valued offset arrays from the same command remain visible, consistent with the byte-only matcher.

Local evidence retained by the reporter: `bench-harbor/health-review/interleaved-vigenere.json`, `reproduce-findings.mts`, and `reproduction-results.json`. The minimal reproduction below is self-contained and does not require these private/local files or benchmark data.

**Limits of this finding:** the saved records establish false redaction, not what the live provider saw on its first request. The cipher task later terminated after an unrelated cleanup command killed its own controlling processes. This issue does not attribute that failure or any benchmark score to redaction.

## Exact cause and affected paths

1. [`sanitizer.ts:352–360`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/secrets/sanitizer.ts#L352-L360) installs an unconditional regex in `SECRET_PATTERNS`. One byte followed by `{31,199}` additional bytes means **32–200 elements**, despite the comment motivating only 32/64-byte key material. Its claim that this excludes ordinary numeric arrays is incorrect. No field name, file provenance, surrounding context, or key-format validation participates in this rule.
2. [`redactSecrets`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/secrets/sanitizer.ts#L397-L417) applies every pattern to every string. [`redactSecretsInValue`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/secrets/sanitizer.ts#L421-L449) recursively applies it to string leaves. A genuine numeric array under an ordinary object key is preserved; the same array printed inside a tool-output string is removed. This issue concerns that textual representation, not all numeric JSON arrays.
3. [`llmMessageToDurableResponseItem`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/session/message-history-conversion.ts#L72-L83) and [`redactResponseItemForPersistence`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/session/message-history-conversion.ts#L260-L339) sanitize durable messages. Tool-result integrity is checked first, then deliberately rebound to the redacted body with representation `redacted`. The altered body passes integrity validation. **This is a false classification and semantic data-loss bug, not evidence of an integrity-check bypass or corrupted digest.** Digest metadata cannot recover the deleted values.
4. The [`rollout serializer`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/session/rollout-item.ts#L320-L328) also sanitizes ordinary event/response records. Compaction payloads are sanitized before their byte counts and hashes are computed in [`createCompactionPayloadBundleV1`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/services/compact/payload-manifest.ts#L30-L45), so a digest-valid payload can already contain the false-redaction marker.

## Persistence, live context, and compaction distinction

- **Fresh in-memory messages:** [`persistNewResponseItems`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/session/run-turn.ts#L2084-L2113) retains the original message content while copying only persisted-integrity metadata from the durable item. The sanitizer conversion does not mutate its input. Therefore saved `[REDACTED_SECRET]` markers alone do not prove that the first provider request lost the arrays.
- **History reconstructed from durable items:** [`responseItemToLlmMessage`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/session/message-history-conversion.ts#L113-L130) copies the already-redacted content back into a model message. Pure-helper round-trip reproduction confirms the marker survives; it cannot reconstruct the original numbers.
- **Canonical compaction:** [`RolloutStore.prepareSource`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/session/rollout-store.ts#L1024-L1067) reads active history from the canonical rollout. [`createAuthoritativeSelectionMapper`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/services/compact/transaction.ts#L1539-L1547) explicitly selects canonical saved bodies; its [`return path`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/services/compact/transaction.ts#L1642-L1648) returns those bodies to the summary plan. Thus false-redacted canonical evidence can become compaction input even while the process stays alive. Legacy replacement conversion also applies the same sanitizer. Post-compaction history is reconstructed from the replacement records in [`run-turn-compaction.ts`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/src/session/run-turn-compaction.ts#L540-L545).

The compaction propagation above is established by source tracing and payload/conversion helper tests. No provider call or end-to-end resume/compaction run was performed for this investigation, and occurrence in this particular benchmark's provider context remains unproven.

## Safe minimal reproduction

From a checkout with installed dependencies and the repository-required Node.js 26.5 toolchain:

```bash
node --import tsx --input-type=module <<'JS'
import assert from "node:assert/strict";
import { redactSecrets, redactSecretsInValue } from "./runtime/src/secrets/sanitizer.ts";

const numbers = Array.from({ length: 80 }, (_, i) => i);
const input = "permutation = " + JSON.stringify(numbers);
const output = redactSecrets(input);
assert.equal(output, "permutation = [REDACTED_SECRET]"); // current bug
assert.equal(redactSecretsInValue({ stdout: input }).stdout, output);
assert.deepEqual(redactSecretsInValue({ values: numbers }), { values: numbers });
console.log(output);
JS
```

Actual: `permutation = [REDACTED_SECRET]`. Expected: this explicitly benign 80-index permutation remains intact. This reproduction uses generated non-secret numbers, performs no network requests, and does not open a real session or modify benchmark artifacts.

Additional isolated helper checks passed against **both commits**: boundaries 31/32/33/64/65/80/100/199/200/201; compact/spaced/multiline lists; negative/256/fraction controls; nested string versus structured numeric arrays; idempotence; unchanged input message; redacted durable body with valid integrity; response-item JSON round trip; replacement-history conversion; compaction payload construction and reconstruction. These checks characterize the bug, not a fix.

## Required fix scope and safeguards

- Replace the blanket byte-list assumption with a documented classification policy that preserves demonstrably benign analytical output while protecting actual credential material at all affected persistence/export sinks.
- Treat raw 32/64-byte lists as **ambiguous**, not inherently safe or inherently a proven key. Exact length alone does not solve this; a benign list and a secret can have identical numeric syntax. Account for available trusted provenance, sensitive fields/key-file origins, supported key formats, and missing context. If an interface needs additional metadata, define how it reaches the sanitizer and how untrusted tool text cannot grant itself an exemption.
- Preserve existing wallet-key, mnemonic, PEM, passphrase, API-token, and other credential protections. Do not disable redaction, add a benchmark-specific exception, whitelist arbitrary tool output, or rely on a `permutation` label alone as trusted proof.
- Preserve tool-result identity, redaction idempotence, and compaction hashing invariants. Do not repair this by weakening integrity checks or re-sanitizing content-addressed chunks after hashing.
- Define handling of already-written markers: original values are absent from those records and cannot be restored from a digest. Any recovery should require a real surviving source, not fabricated values or silent history rewriting.

## Acceptance criteria and regression coverage

- [ ] The benign 80/100-element benchmark-shaped permutation examples remain byte-for-byte intact in string sanitizer output, nested `stdout`/`result` strings, durable response items, serialized rollout events, and content-addressed compaction payloads.
- [ ] Test benign permutations/counts/byte data near 31/32/33/63/64/65/79/80/100/199/200/201 elements; compact, spaced, and multiline forms; negative, out-of-range, and fractional controls; and numeric JSON arrays versus embedded text. For ambiguous 32/64-byte examples, assert the documented context/provenance policy rather than an unconditional length exemption.
- [ ] Actual credential fixtures remain redacted, including supported 32-byte secret and 64-byte keypair representations with appropriate provenance/context, multiline key-file output, nested sensitive fields, and other existing sanitizer families. Add adversarial cases so cosmetic labels or whitespace cannot bypass real-key protection. Use generated/synthetic fixtures, never production credentials.
- [ ] A hermetic end-to-end session test proves fresh provider-facing content, durable serialization, reconstructed/resumed history, and compaction selection agree with the intended benign-data policy; secret fixtures remain protected at the durable/export boundary.
- [ ] Tool-result original identity and persisted representation remain valid, and compaction chunks round-trip with correct hashes. Existing tampering-rejection and secret-redaction tests still pass.
- [ ] Correct the comment claiming the current byte range excludes ordinary numeric arrays, and document the unavoidable ambiguity and chosen policy.

The existing [`sanitizer-wallet-c1.test.ts`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/tests/secrets/sanitizer-wallet-c1.test.ts#L36-L43) covers intended wallet redaction, but its [`does not over-redact ordinary integer arrays`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/tests/secrets/sanitizer-wallet-c1.test.ts#L82-L87) test only uses very short or out-of-range lists, so it misses this case. [`a3b-tool-result-cutover.test.ts`](https://github.com/tetsuo-ai/agenc-core/blob/f037754766c3f0ad543f85a3c3db1eea7afa3b01/runtime/tests/session/a3b-tool-result-cutover.test.ts#L138-L168) covers valid intentional secret redaction; add benign evidence-preservation coverage without deleting that protection.

turkonthelurk:
Fixes https://github.com/tetsuo-ai/agenc-core/issues/2476

## Problem

`runtime/src/secrets/sanitizer.ts` installed an unconditional regex that redacted every textual bracketed list of 32–200 comma-separated bytes as a Solana `id.json` keypair, and its comment claimed this excluded ordinary numeric arrays. It did not. An `exec_command` result printing `T[:100]` / `invT[:80]` index permutations reached the rollout as `[REDACTED_SECRET]`, and resume, checkpoint reconstruction and compaction source selection all read the marker instead of the evidence, while the same numbers as a structured JSON numeric array were preserved.

The hypothesis in the issue held up: the unconditional 32–200 byte-list regex is the only false-positive source; the persistence sinks (`llmMessageToDurableResponseItem`, `serializeRolloutItem`, `createCompactionPayloadBundleV1`) behave correctly given a correct sanitizer, so none of them changed and no integrity/hashing invariant was touched.

## Fix: a documented classification policy

The regex is replaced by `redactIntegerListKeyMaterial`, a single linear-pass scanner with an explicit policy (the full rationale is in the doc comment above `AMBIGUOUS_RAW_KEY_LENGTHS`):

1. **Supported key formats.** Only exactly 32 (secret scalar) or 64 (keypair) unsigned bytes are raw ed25519 encodings. A list of any other length is not a supported raw-key encoding and is **preserved byte-for-byte**. Length is a *negative* test only; it never proves a list is a key.
2. **Missing context → fail closed.** A 32/64-element unsigned-byte list is ambiguous (a permutation of 0..31 and a secret scalar are the same text) and is **redacted**. Nothing in the surrounding text can lift this: `permutation =`, `T[:64]`, `# benign:` etc. are untrusted tool output, so a cosmetic label cannot grant a real key an exemption. Trailing commas and odd whitespace do not dodge the rule.
3. **Sensitive-field / key-file context → widen only.** An integer list of any length or value range directly labelled by a sensitive key name (`"secretKey": [...]`, `private_key = [...]`, `SEED_PHRASE=[...]`, via the shared `isSensitiveKey`) is redacted, mirroring the structured leaf rule. In-band context can only widen redaction, never narrow it.
4. **Trusted provenance is structural only.** In `redactSecretsInValue` a numeric JSON array is runtime-typed data, preserved under an ordinary key and leaf-redacted under a sensitive key. That decision is made from the JSON position, never from string content, so no tool can reach it by formatting stdout. No new text-side "trust me" interface is introduced, because there is no caller that could legitimately vouch for tool text.

Signed-byte dumps, fractional values and values > 255 are never key material under rules 1–2; rule 3 still redacts them under a sensitive label.

Already-written `[REDACTED_SECRET]` markers are not recoverable (the durable record holds only a digest) and are not rewritten; this is stated in the policy comment.

Not done, deliberately: no redaction disabled, no benchmark-specific exception, no whitelist of tool output, no change to integrity checks or to content-addressed chunk handling.

## Tests

New `runtime/tests/secrets/sanitizer-integer-lists.test.ts` (27 tests; 16 fail against the old sanitizer) covering issue reproduction, boundaries, adversarial labels, JSON-vs-text, sensitive labels, multiline key-file output, and idempotence.

New hermetic session test `runtime/tests/session/benign-integer-arrays-durable-boundary.test.ts` driving benign permutations and a synthetic keypair through durable items, rollout, reconstruction, compaction (digest-verified), and cold reopen.

### Ran

27 related test files / 369 tests passed; `tsc` clean for src and test-support.

## Notes for review

- A longer or non-byte list under a sensitive *text* label is now redacted too (rule 3 widen-only). Structured behaviour is unchanged.
- Sibling-test filler composed of BIP39 words is redacted by the bare-mnemonic detector (pre-existing, unrelated).
- Possible follow-up: a structured 64-number array whose second half is the ed25519 pubkey of the first half could be redacted with zero false-positive risk.

Opened from fork `turkonthelurk/agenc-core` (Cursor cannot install the GitHub App on `tetsuo-ai` yet). Companion fork PR: https://github.com/turkonthelurk/agenc-core/pull/1


signerless:
Superseded by #2554, which keeps this PR's integer-list policy intact (length as a negative test, fail-closed on ambiguous 32 and 64 element lists, sensitive-label context widening only, structural provenance as the only trusted signal) and fixes the one defect in it.

The hand-rolled bracket scanner's isWhitespaceCode accepted only six ASCII codepoints while the rule it replaced used an ECMAScript \s class, so a byte list formatted with NBSP, EM SPACE, LINE SEPARATOR or NARROW NBSP never parsed as 32 or 64 elements and reached durable tool history unredacted. #2554 extends the predicate to the full \s set and adds a regression test covering both separators and the benign 80 element array from #2476.

PR #2493, opened 2026-09-15
feat(runtime): isolated execution environment groundwork

7etsuo:
## Problem

Terminal-Bench trials can kill AgenC's own CLI and process brokers because task commands run in the controller's PID namespace and see task filenames in controller argv (#2477). The runtime has no explicit execution environment: file tools, configuration, instructions and process launches all use the controller filesystem and host spawns.

## Resulting behavior

This lands the groundwork recorded in `docs/design/execution-environment-implementation.md` (the implementation ledger, which is the acceptance contract and lists every remaining item):

- Host supervisor and additive Docker runtime adapter (`runtime/native/execution-host/`), one-use command leases, host-owned cgroups, receipts and output cursors.
- Immutable native filesystem worker (`agenc-filesystem-worker.c`, protocol 9) with openat2 task-root containment, bound reads, transaction guards, directory mutations and exclusive recursive parent creation with partial-creation evidence.
- `runtime/src/execution/`: Docker process/filesystem/ripgrep adapters, environment binding, permission path evidence, coordinator paths.
- Environment-scoped unified-exec manager with durable numeric handles, owner-scoped `list_processes`, controller-crash recovery; checkpoint v5 / rollout schema v6.
- Configuration, instructions, rules, plugins, skills, roles, agent memory prompts, permission evaluation, FileRead, Write, Edit and MultiEdit route through the selected task filesystem when an execution workspace is bound.
- Isolated bootstrap and the CLI environment selector remain disabled; local execution is unchanged for existing sessions.

Still pending per the ledger: apply_patch/NotebookEdit branches, editor overlays, LSP feedback, early bootstrap, durable run recovery integration, installer/preflight, Harbor adapter, final acceptance gates (E3, F2 limits, M2, M3, B1, D1, V1–V3).

## Validation (rebased on origin/main at 2b38daea0)

From the repository root with `PATH=/tmp/agenc-node-26.5.0/bin:$PATH`:

- `npm run typecheck` (runtime) — production and test-support, exit 0
- `npm run build` (runtime) — build, package entrypoints, generated SDK types, exit 0
[...]

Issue #2477, opened 2026-09-13
[URGENT] Harden AgenC against accidental self-termination during process cleanup


7etsuo:
## Problem and urgency

AgenC aborted a nearly 2 h 48 min Terminal-Bench trial when its generated cleanup command killed its own CLI and two process brokers. The command searched `/proc/*/cmdline` for task filenames; AgenC exposed those filenames in its **prompt argv**, and the native brokers exposed them in their **shell-command argv**. Excluding only the cleanup Python process did not protect the controlling processes.

**Priority: urgent. Assignee: @7etsuo.** This is a confirmed agent-induced self-termination and an actionable runtime/harness reliability gap. It is **not evidence of a sandbox escape**: this run explicitly used `--dangerously-bypass-approvals-and-sandbox --add-dir /`. We should remove the avoidable argv collisions and make managed-process cleanup reliable and easy to recover; we must not imply that a JavaScript command filter can block arbitrary same-UID/root Python signals.

## Reproduction environment and observed result

- Terminal-Bench 4.0, `interleaved-vigenere__CoCziJ9`, job `tb4-grok-cec48942cfa9-clean`; Docker Linux amd64, `grok/grok-4.6`, high effort, one attempt.
- Running source build: `cec48942cfa97fd095f057779d8c55b6cb39a7f3`, despite its package version still reading `0.17.0`. Current code reviewed: `f037754766c3f0ad543f85a3c3db1eea7afa3b01`.
- Failure recorded at **2026-09-13 21:53:55 UTC**. Harbor reported `NonZeroAgentExitCodeError`, **exit 137**, and the shell explicitly reported PID **568** (`agenc ...`) as `Killed`.
- Cleanup call `call-6684f372-f0f1-4958-abb7-ae8731d3fbc1-105` enumerated command lines, matched either `cracker.py` or `sample_ciphertext`, and called `os.kill(pid, signal.SIGKILL)`, skipping only `os.getpid()`.
- The completed tool output listed killed PIDs **568** (CLI), **9446** and **10230** (brokers), and **9447** and **10231** (shells). The Python cleanup PID was **10232**.
- AgenC then reported: `AgenC could not verify descendant process cleanup: exec_command process 88 Linux subreaper cleanup could not be verified: Linux process containment broker exited without cleanup proof`.
[...]

PR #2549, opened 2026-09-17
fix(runtime): recover background work through owned process identity

turkonthelurk:
Refs https://github.com/tetsuo-ai/agenc-core/issues/2477 — **remaining** scope after #2494.

#2494 already landed the direct argv-collision fix (prompt over stdin, broker bootstrap over a private FD). This PR covers only the two items the maintainer left open: recovery through owned process identity, and an honest statement of the protection boundary. #2493 (isolated execution environment) is not resurrected.

## Problem

An agent recovering from hung/background work matched task filenames against `/proc/*/cmdline` and SIGKILLed the AgenC CLI and two process brokers along with its work. The manager already knew every yielded session and its owner; the model had no owner-scoped way to ask which of its sessions are still live after a `kill_process` returned `terminated:false`.

## Changes

**Process manager**
- `listOwnedProcesses({ ownerId })` — yielded sessions of one owner only; never another owner's; never a process-table scan. `stopping` means signalled but exit not yet observed.
- `terminateOwnedProcesses({ ownerId, processIds? })` — bulk stop through manager identities. Named ids are ownership-checked in full before any signal (`owner_denied` refuses the whole batch with no effect); unknown/exited ids report `terminated:false`.
- `isProcessOwnedBy` — enumeration predicate stricter than per-id access.

**Tool surface**
- `kill_process` accepts `session_id`, `session_ids`, or `all: true`. Every result carries `owned_live_sessions` so `terminated:false` points at remaining owned work instead of inviting a scan.
- New read-only `list_processes` tool, registered and available to read-only delegation.
- System prompt states the managed cancellation path and why a cmdline scan also selects AgenC's own control processes.

**Docs**
- "Recovering background work" section plus explicit "What this does not guarantee": argv hygiene and managed cancellation are reliability mitigations, not an OS-enforced boundary under `danger-full-access`.
- Fail-closed lost-cleanup-proof diagnostic deliberately preserved.

## Verification

`tsc --noEmit` clean. Focused unified-exec / kill-process / list-processes / supervisedProcess / related suites green. Broader sweep failures match unmodified `main` in the agent environment (pre-existing).

## Not claimed

Does not prevent same-UID SIGKILL of the controller without OS-enforced isolation (#2493 out of scope).

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/2


signerless:
Superseded by #2555, which keeps the stopping status and the ownership model introduced here and fixes the two readers that ignored it.

list_processes filtered the listing on running or stopping but computed live_count on running alone, so this PR's own test asserted a three row listing alongside live_count 2. kill_process built owned_live_sessions the same way, so a session it had just signalled dropped out of the remaining work before its exit was observed. #2555 adds a single isLiveOwnedProcess predicate used by all three sites, corrects that assertion to 3, and adds a regression test that drives a real TERM trapping process through every selector shape.

Issue #2520, opened 2026-09-15
Auto-compaction never fires in long tool-call turns: the conservative-fallback estimate walks the output reservation to zero and the turn dies at the context window


signerless:
## What happens

A long tool-call turn on a provider without a native tokenizer runs until the adapter refuses its own request at the context window, and auto-compaction never fires once. The turn dies with `turn_execution_failed` and the CLI exits 1.

Observed on a 4 h 50 min Terminal-Bench trial (GLM-5.3 on zai-coding-plan, core `6f0785647`, 826 tool iterations in a single turn, 782,598 rollout events):

```
turn_failed: zai-coding-plan error: estimated prompt (948744) plus reserved output (1256)
             exceeds context window (950000) after 1024 token safety buffer
run_terminal: status=cancelled, stopReason=one_shot_complete
```

## Evidence

1. Every one of the 830 model dispatches used `tokenAccountingSource: conservative_fallback` with `tokenAccountingConfidence: conservative` (no native tokenizer for this provider), so the input figure is byte-derived. The provider's own reported `promptTokens` for the same history peaked at 281,946, about 3.4x lower than the 948,744 the refusal used.

2. `fitRequestWithinContextWindow` (`runtime/src/llm/providers/openai/adapter.ts`) absorbed the growth silently by shrinking the output reservation. `maxOutputTokens` per dispatch:

   | time (UTC) | maxOutputTokens | implied input estimate |
   | --- | --- | --- |
   | 18:33 to 22:03 | 131,072 | below 818k |
   | 22:11 | 113,196 | about 836k |
   | 22:36 | 77,848 | about 871k |
   | 22:57 | 40,774 | about 908k |
   | 23:11 | 21,084 | about 928k |
   | 23:20 | 4,358 | about 945k |
   | 23:23 | 1,256 | 948,744, then the throw |

   The estimate crossed the auto-compaction threshold (0.75 x 950,000 = 712,500, `AUTOCOMPACT_MAX_WINDOW_FRACTION`) before 22:03, and the run continued for another 80 minutes and roughly 110 dispatches.

3. Auto-compaction never ran. Across all 782,598 rollout lines there is not a single event mentioning compact or compaction: no `auto_compact_deferred`, no `compact_failed`, no pre-sampling compaction failure. `AGENC_DISABLE_AUTO_COMPACT` was not set, so the limit was finite.

[...]

PR #2552, opened 2026-09-17
fix(recovery): walk compaction ladder on 413 collapse and warn on squeezed output (#2520 slice A)

turkonthelurk:
Refs https://github.com/tetsuo-ai/agenc-core/issues/2520 — **slice A only** (recovery retry restoration + squeeze warning).

## Summary

1. **Recovery retry restoration** (distinct from #2523 classification). The reactive 413/overflow collapse now walks the degraded compaction ladder instead of ending the turn on the first failure.
2. **Silent squeeze warning.** `fitRequestWithinContextWindow` emits when the granted output reservation falls below half of the requested maximum.

Addresses the incident facts: recovery threw while unused ladder tiers remained, and the output reservation decayed for ~80 minutes with nothing visible in the rollout.

## Not redone (already merged by @signerless)

- #2521 planner splits source at encoding limits
- #2523 compaction resource refusal classified as refusal, not trigger fault
- #2524 pressure gate charges retained provider reasoning fields
- #2526 compaction source digest uses its own budget

## Out of scope

- Conservative-fallback bytes/token divisor calibration (needs replay/measurement first)
- Claiming the live proactive gate would have fired on the incident history
- Relaxing untrusted provider-output validation ceilings
- Routing gate and admission through one shared accounting service (slice B)

## Changes

**Reactive recovery** — attempt `standard`, then `aggressive_summary` / `emergency_local` per `nextLadderTiers`, sharing episode ladder state with the proactive path. Compaction transaction / cannot-reduce failures that leave history unchanged are tier failures and step down (with `auto_compact_degraded`). When all tiers decline: `context_collapse_ladder_exhausted` then typed `prompt_too_long_exhausted` from #2523 — no untyped `recovery_trigger_threw`. Fatal/uncertain failures still propagate as trigger faults.

Note for review: `output_schema_invalid` / `provenance_invalid` from a tier now step the ladder (matching proactive `ladderAppliesToDecline`) rather than immediately faulting; flagged deliberately vs a strict reading of #2523.

**Squeeze warning** — below 50% of requested output tokens, emit `output_reservation_squeezed` (re-arm on fit; warn again each time reservation halves). Fit/admission math unchanged.

## Verification

`typecheck` clean. New tests fail on `main`, pass with the fix. Focused phases/recovery/compact/ladder/adapter suites: 68 files / 907 passed (includes #2521/#2523/#2524/#2526 coverage). Broader session/budget/llm: 4758 passed; one unrelated cron-timing flake also seen on pristine `main`.

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/3


signerless:
nice, don't merge yet; check back after my benchmark finishes

signerless:
Superseded by #2557, which keeps this PR's compaction-ladder work and the squeeze warning it introduced, and fixes the two gaps that made that warning unreachable in practice.

The warning only fired when the chat-completions adapter shrank the reservation itself. On the admitted path `runAdmittedModelCall` fits the reservation to the context window before dispatch and sends the fitted value, so the adapter compared that value against itself, took its early return and cleared the warning latch. Identical wire bytes, no warning. The incident this warning exists to surface, #2520's decay from 131072 output tokens to about 1200 over roughly 110 dispatches, was the one case it stayed silent for.

A second review then found the same gap one branch over: the shrink branch measured against the admitted value, so a reservation squeezed by admission and then again by the adapter also reported nothing.

#2557 carries the pre-admission ceiling as an internal option that never reaches the wire and never widens a limit, passed only when a squeeze actually occurred, and forwards it from both call sites. The wire ceiling, minimum output, provider buffer and all accounting are untouched.

Closing rather than leaving open, because this head still contains the original implementation and should not be merged after its replacement.

PR #2561, opened 2026-09-18
fix(permissions): honor content-specific ask rules before mode auto-allows (#2125)

turkonthelurk:
Fixes https://github.com/tetsuo-ai/agenc-core/issues/2125

**Draft, do not merge.** Opened from fork `turkonthelurk/agenc-core` for review only. Nothing here merges anything; no push to `main` on either repo.

## Problem

Verified on current upstream `main` (`6a2a92d96`, after #2558 and #2559): a content-specific `ask` rule could be overridden by two mode shortcuts.

1. `isPathAllowed` matched `ask` rules only after the working-directory auto-allow, so a workspace read in any mode and a workspace write under `acceptEdits` returned `{ type: "mode" }` and the rule never surfaced.
2. `checkToolPathPermission` converted every non-deny, non-safety rejection into an allow under `bypassPermissions`. A matched ask rule was included unless it named the session plan file.

Settings could say a file requires confirmation while the runtime read or edited it automatically. The evaluator already treats a rule-based ask reported by the tool as bypass-immune (step 1f), but path validation never reported it in these cases.

## Changes

`runtime/src/permissions/path-validation.ts`

- The `ask` rule lookup moves directly under `deny` and the protected-path safety floor, above the working-directory and mode auto-allows. Order is now: deny, plan-file / durable-memory carve-outs (which already honored ask), safety check on writes, **ask**, working-directory auto-allow, allow, working-directory reject.
- The `bypassPermissions` shortcut in `checkToolPathPermission` no longer converts any rule-based ask into an allow; it lifts only the working-directory prompt. The plan-file special case collapses into the general rule.
- `deny` still outranks `ask`; the safety check still outranks both (a `Write` ask on `.git/config` still returns `safetyCheck`, fail-closed); `ask` still outranks a conflicting `allow` regardless of source.

`docs/reference/tools-permissions-sandbox.md`: one paragraph stating the path-rule order and that `bypassPermissions` does not lift a rule-based ask.

[...]

Issue #2125, opened 2026-09-04
fix(permissions): honor content-specific ask rules before mode auto-allows


7etsuo:
## Summary

Content-specific `ask` rules can be overridden by two mode shortcuts. They are checked after the working-directory auto-allow, and the tool wrapper converts a matched ask outside the workspace into an allow under `bypassPermissions`.

## Verified current behavior

[`isPathAllowed`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/permissions/path-validation.ts#L357-L449) applies rules in this order:

1. `deny`
2. path safety
3. working-directory/mode auto-allow
4. `ask`
5. `allow`

[`checkToolPathPermission`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/permissions/path-validation.ts#L672-L695) then auto-allows every non-deny, non-safety rejection in bypass mode. A matched content-specific ask has `decisionReason.type === "rule"`, so it is included in that shortcut.

On the audited commit, a direct reproduction added an exact `FileRead` ask rule for `runtime/package.json`. `validatePath` returned:

```json
{
  "allowed": true,
  "decisionReason": { "type": "mode", "mode": "default" }
}
```

An exact `Write` ask rule for a workspace path under `acceptEdits` likewise returned `allowed: true` with the mode as its reason.

A second reproduction added an exact `FileRead(/etc/hosts)` ask rule under `bypassPermissions`. The path matcher found the rule, but `checkToolPathPermission` returned `behavior: "allow"` with `bypassPermissions` as the reason.

The evaluator treats a content-specific ask as bypass-immune when the tool reports it in [`evaluator.ts`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/permissions/evaluator.ts#L338-L345), but path validation never reports the rule in these common cases.

## Impact

User, project, and managed settings can state that a particular file requires confirmation while the runtime reads or edits it automatically. The behavior conflicts with the displayed permission policy and removes a requested review boundary.

## Fix direction

Evaluate content-specific `ask` rules before working-directory and mode auto-allows, immediately after `deny` and mandatory safety checks. Do not convert a rule-based ask into a bypass allow. Preserve documented precedence when multiple sources define conflicting rules.

## Acceptance criteria

[...]

PR #2562, opened 2026-09-18
fix(permissions): match path rules using filesystem case semantics (#2126)

### LINKED ISSUE #2126: fix(permissions): match path rules using filesystem case semantics
## Summary

The current path-rule matcher compares exact paths, recursive prefixes, and generated glob expressions with case-sensitive JavaScript operations. It does not account for filesystems where differently cased names identify the same file.

## Verified current behavior

- [`matchPathRuleContent`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/permissions/path-validation.ts#L262-L273) uses `===`, `startsWith`, and a case-sensitive `RegExp`.
- [`isPathInside`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/permissions/path-validation.ts#L127-L133) also compares normalized path strings without filesystem case handling.
- The older permission utility documents and provides [`normalizeCaseForComparison`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/utils/permissions/filesystem.ts#L92-L103) specifically to prevent mixed-case permission-check mismatches.

There is no platform or filesystem case branch in the new matcher. On default Windows and commonly configured macOS volumes, a rule for `C:/Work/Secret.txt` can miss the same file addressed as `c:/work/SECRET.txt`.

## Impact

Allow, ask, and deny rules can disagree with filesystem identity. A deny or ask policy may not apply to the file that the operating system opens, while an allow policy may unexpectedly prompt or fail.

## Fix direction

Normalize both rule patterns and candidate paths according to the target filesystem's case semantics before exact, prefix, and glob matching. Preserve the original strings for display and audit records. Avoid lowercasing on known case-sensitive volumes if runtime detection is available.

## Acceptance criteria

- Exact, recursive, and glob rules match alternate casing on case-insensitive filesystems.
- Case-distinct files remain distinct on case-sensitive filesystems.
- Drive-letter and separator normalization is covered on Windows.
- Tests cover `allow`, `ask`, and `deny` behavior for mixed-case forms.

## Duplicate check

GitHub issue and pull-request searches for case-insensitive path rules, Windows permission casing, and mixed-case policy paths found no matching report.

turkonthelurk:
Fixes https://github.com/tetsuo-ai/agenc-core/issues/2126

**Draft, do not merge.** Opened from fork `turkonthelurk/agenc-core` for review only. Nothing merges; no push to `main` on either repo.

## Problem

`matchPathRuleContent` compared exact paths, recursive `/**` prefixes and generated globs with case-sensitive JavaScript (`===`, `startsWith`, case-sensitive `RegExp`), and `isPathInside` compared working-directory containment the same way. On default Windows and macOS volumes, where `Secret.txt` and `SECRET.txt` are one file, a rule for `C:/Work/Secret.txt` missed the same file addressed as `c:/work/SECRET.txt`. Allow, ask and deny could therefore disagree with the file the OS actually opens.

## What changed

- **New `runtime/src/permissions/path-case.ts`**
  - `pathCaseSemantics(path)` returns `"sensitive" | "insensitive"` for the volume holding a path. It walks to the nearest existing entry, stats it under its original and case-flipped spelling, and compares `dev`/`ino`. Verdicts are cached per containing directory (bounded), so siblings and files created later cost no further syscalls. When nothing on the path can be probed (a bare root, a drive root) it falls back to the platform default (Windows/macOS insensitive, everything else sensitive). This is what lets a case-sensitive APFS volume or a case-insensitive NTFS/SMB mount on Linux be handled correctly rather than by platform alone.
  - `comparablePath(path, semantics)` folds case for comparison only, reusing the existing `normalizeCaseForComparison` from `protected-paths.ts` (the `utils/permissions/filesystem.ts` copy is behind the donor-stack import boundary, so it is not imported).
  - `__setPathCaseSemanticsResolverForTesting` lets tests exercise both semantics on one host.
- **`runtime/src/permissions/path-validation.ts`**
  - `matchPathRuleContent` folds both rule and candidate under the candidate's semantics before the exact, `/**` prefix and glob checks. Separators were already folded to `/`, so a Windows drive letter and mixed separators (`C:\Work\Secret.txt` vs `c:/work/SECRET.txt`) now match together.
  - `isPathInside` folds the same way, so a working root or memory root spelled differently from the candidate still contains it where the filesystem says it is the same directory.
  - `matchingRuleForPath` resolves semantics once per candidate instead of once per rule x candidate.
  - Rule text and the `PermissionRule` returned in `decisionReason` are untouched; only the comparison folds.
- **`runtime/knip.config.mjs`**: lists the new module alongside the other permission files whose test-only exports the production graph cannot see.
- **`docs/reference/tools-permissions-sandbox.md`**: one paragraph on the semantics.

Out of scope, deliberately: #2125 ask-rule-vs-mode precedence (draft #2561) and the merged #2558/#2559 work; this branch is based on upstream `main` at `6a2a92d96` and does not depend on #2561.

## Tests

New `runtime/tests/permissions/path-rule-case-semantics.test.ts` (35 tests):

- `matchPathRuleContent` with forced semantics: exact / `/**` / `*` / `**` / `?` rules, backslash rule and backslash target, drive-letter casing. Each case matches under `insensitive` and stays distinct under `sensitive`. Folding never widens a rule past its own subtree (`C:/Work/**` does not match `c:/workspace/...`).
- `pathCaseSemantics` against the real temp volume: an existing file, a not-yet-created deep path (inherits from its ancestor), a directory that holds `same.txt` and `SAME.TXT` as distinct files, and the platform-default fallback. These assert against what the host volume actually does, so they pass on both case-sensitive and case-insensitive hosts.
- End-to-end through `checkToolPathPermission` for **allow**, **ask** and **deny** (session rules; absolute exact, absolute `/**`, relative `./Secret/**`, glob) with `Secret/Notes.txt` addressed as `sECRET/nOTES.TXT`:
  - forced insensitive: deny denies, ask asks with `decisionReason.type === "rule"`, allow allows, a working directory contains its own recased spelling, and the matched rule keeps its original `ruleContent`.
  - forced sensitive: the recased path is not matched by deny/ask/allow and falls through to the mode / working-directory outcome; the exact spelling is still denied.

Before the `path-validation.ts` change, 17 of those 35 fail (every insensitive-side case); after, all pass.

Also run locally (Node 26.5.0):

- `tsc --noEmit` on `runtime`: clean.
- Permissions + file-tool + agents suites: 2262 passed; 11 unrelated failures reproduce on pristine upstream `main` on Linux.
- Diff-scoped fast gate passed aside from one timing-flaky `run-turn` cron test that passes in isolation with and without this change.

## Merge policy

Draft only. Nothing pushed to `main` on either repository.

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/5


7etsuo:
- [path-case.ts:47](https://github.com/tetsuo-ai/agenc-core/blob/b7dc1a7713e52b95a123e1561fcef1a5af629cff/runtime/src/permissions/path-case.ts#L47) uses Unicode case mapping: on NTFS `ß.txt` probes as `SS.TXT`, hits ENOENT, and [path-case.ts:88](https://github.com/tetsuo-ai/agenc-core/blob/b7dc1a7713e52b95a123e1561fcef1a5af629cff/runtime/src/permissions/path-case.ts#L88) caches "sensitive" process-wide, disabling the new deny matching for that directory. Flip only ASCII letters; don't cache ENOENT verdicts.
- [path-case.ts:82](https://github.com/tetsuo-ai/agenc-core/blob/b7dc1a7713e52b95a123e1561fcef1a5af629cff/runtime/src/permissions/path-case.ts#L82) follows links: a case-twin symlink or hardlink makes a case-sensitive directory read "insensitive". Use `lstat` and check the directory entry, not just the inode.
- [path-case.ts:134](https://github.com/tetsuo-ai/agenc-core/blob/b7dc1a7713e52b95a123e1561fcef1a5af629cff/runtime/src/permissions/path-case.ts#L134) folds the whole path, so a new entry under a case-sensitive mount root inherits the parent volume's "insensitive" verdict, widening allow rules. Fold only the probed volume or canonicalize with `realpathSync.native`.
- Add regressions for each; land after #2561 (docs conflict).

PR #2564, opened 2026-09-18
fix(sdk): wait for subprocess stdout to drain before settling results (#2091)

turkonthelurk:
Closes https://github.com/tetsuo-ai/agenc-core/issues/2091

**Draft for review only.** Not to be merged by automation; a maintainer decides when and whether it lands. Opened from fork `turkonthelurk/agenc-core`. Nothing pushed to `main` on either repo. Independent of draft #2563 (#2090).

## Problem

The SDK subprocess transport settled a run as soon as the child emitted `exit`. Node can emit `exit` before stdout closes, so a valid stream-json result still in the pipe (or written by an unreferenced descendant that inherited stdout) arrived after the SDK had already rejected the run with `AgenC CLI exited (code 0) without a stream-json result`.

Verified still present on `tetsuo-ai/agenc-core@6a2a92d` (`main` at the time of this PR) before changing anything. No open or closed PR referenced #2091.

## Change

- **`AgencSubprocessChild`** now exposes stdout `end`, child `close`, `on`/`removeListener` for `error`/`exit`/`close`, optional `pid`, and optional stdio `destroy`.
- **`promptViaSubprocess`** records process-exit state separately from stream-drain state. It parses complete lines as they arrive, parses the final unterminated line **once** after stdout ends, and settles only after both `exit` and `close`.
- Spawn error, abort, exit, stdout end, and close handlers are idempotent and remove every listener on settle.
- **`postExitDrainTimeoutMs`** (default `DEFAULT_POST_EXIT_DRAIN_TIMEOUT_MS` = 5s) bounds a hung post-exit drain. Timeout produces a distinct error and SIGKILLs the child plus, on Unix, its process group. Default Unix spawn uses `detached: true` so that group kill can reach retained descendants.
- Custom spawn adapters must emit `end` and `close` after stdio actually finishes; a Node `ChildProcess` already does.

## Tests

Hermetic, fail-before / pass-after in `runtime/tests/sdk-package/subprocess-transport.test.ts`:

- Valid result after `exit` and before `close` resolves successfully (fake + real wrapper/descendant).
- Final unterminated line is parsed once after stdout `end`; later chunks are ignored.
- Child that closes without a result still returns the current missing-result error.
- Abort and spawn-error races settle once and leave zero leftover listeners.
- Post-exit drain timeout is a distinct error and calls `SIGKILL`.

```text
cd runtime && node scripts/run-hermetic-vitest.mjs run tests/sdk-package/subprocess-transport.test.ts
# 13 passed
[...]

Issue #2091, opened 2026-09-04
fix(sdk): wait for subprocess stdout to drain before settling results


7etsuo:
## Summary

The SDK subprocess transport decides whether a run produced a result as soon as the child emits `exit`. Node can emit `exit` before stdout closes, so a valid result still in the pipe can arrive after the SDK has already rejected the run.

Audited commit: `6d9243b2820821ac0443eaabee58fc89d156b688`

## Current behavior

[`AgencSubprocessChild`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/packages/agenc-sdk/src/subprocess.ts#L35-L55) exposes child `exit` but not child `close` or stdout `end` in the injected transport contract.

The handler at [`subprocess.ts` lines 238-270](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/packages/agenc-sdk/src/subprocess.ts#L238-L270) parses the current remainder and settles the run immediately on `exit`.

Node's child-process contract permits `exit` before stdio streams close. The `close` event is the event that follows completed stdio closure.

## Reproduction

1. Run the transport against a wrapper child that starts an unreferenced descendant with inherited stdout.
2. Let the wrapper exit successfully.
3. Have the descendant write one valid stream-JSON result line to the inherited pipe shortly afterward, then close it.

The SDK rejects early with `AgenC CLI exited (code 0) without a stream-json result`. The valid result arrives after settlement and is ignored.

## Expected behavior

The transport should record the process exit status but keep parsing until stdout has ended and the child has closed. It should parse the final partial line once, then make the result decision.

## Impact

Delayed pipe delivery, high output, or a descendant that briefly retains stdout can turn a valid successful run into a false failure. Tail events and stderr can also be omitted.

## Suggested fix

- Extend the child abstraction to expose stdout completion and child `close`.
- Separate process-exit state from stream-drain state.
- Settle only when the required terminal conditions are known, with an optional bounded post-exit drain deadline.
- Make spawn error, abort, exit, stdout end, and close handlers idempotent.

## Acceptance criteria

- A valid result delivered after `exit` but before `close` resolves the run successfully.
- The final unterminated line is parsed once after stdout ends.
- A child that closes without a result still returns the current clear error.
- Abort and spawn-error races settle once and clean every listener.
[...]

Issue #2090, opened 2026-09-04
fix(sdk): report prompt-event queue overflow


7etsuo:
## Summary

Both SDK prompt transports silently discard the oldest queued events after 1,000 entries. Consumers can receive a successful result and an apparently complete event stream even though hundreds of text, tool, status, or audit events were removed locally.

Audited commit: `6d9243b2820821ac0443eaabee58fc89d156b688`

## Current behavior

The socket-backed prompt channel removes old entries with `shift()` in [`client.ts` lines 349-406](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/packages/agenc-sdk/src/client.ts#L349-L406). The subprocess-backed channel does the same in [`subprocess.ts` lines 194-210](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/packages/agenc-sdk/src/subprocess.ts#L194-L210).

The public event union has an `event_gap` type at [`events.ts` lines 82-91](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/packages/agenc-sdk/src/events.ts#L82-L91), but local queue overflow emits no gap or error. The subprocess test at [`subprocess-transport.test.ts` lines 367-399](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/tests/sdk-package/subprocess-transport.test.ts#L367-L399) currently checks only that the queue stays bounded and accepts the silent loss.

## Reproduction

1. Start a prompt run through either public SDK transport.
2. Emit 1,500 valid message events followed by a successful result.
3. Await `run.result()` before consuming the async event iterator.
4. Collect the iterator output.

The consumer receives 1,000 events, starting at event 500 and ending at event 1499. It receives zero gap markers and no overflow error.

## Expected behavior

Bounded memory is correct, but data loss must be visible. The transport should apply backpressure, fail the stream with a typed overflow result, or emit a non-evictable local-overflow marker that identifies the loss.

## Impact

Slow or result-first SDK consumers can persist incomplete transcripts and audit trails while believing the run was complete. Subprocess clients have no daemon replay path that could restore the discarded events.

## Suggested fix

- Define explicit SDK-local overflow semantics for both transports.
[...]

PR #2565, opened 2026-09-18
fix(sdk): enforce a 16 MiB byte limit on subprocess JSON lines (#2092)

### LINKED ISSUE #2092: fix(sdk): enforce a byte limit on subprocess JSON lines
## Summary

The SDK subprocess transport appends stdout into an unbounded string until it finds a newline, then passes each complete line to `JSON.parse`. A child that emits a very large line or never emits a newline can consume unbounded memory and parsing time in the embedding process.

## Current behavior

[`stdoutRemainder`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/packages/agenc-sdk/src/subprocess.ts#L149-L156) is a string with no byte accounting. The data handler at [`subprocess.ts` lines 218-227](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/packages/agenc-sdk/src/subprocess.ts#L218-L227) concatenates every chunk before splitting and parsing lines.

The SDK socket transport already enforces a 16 MiB frame ceiling in [`socket.ts` lines 264-278](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/packages/agenc-sdk/src/socket.ts#L264-L278), but the subprocess path has no equivalent limit.

## Reproduction

1. Provide a child that writes one invalid 17 MiB line followed by a newline.
2. Have it then write a valid result line.
3. Run the normal subprocess decoder.

The decoder allocates and parses the oversized line, ignores it as non-protocol output, and accepts the later result. If the first writer never sends a newline, `stdoutRemainder` can continue growing for the lifetime of the process.

## Expected behavior

The decoder should enforce a documented byte limit incrementally, before concatenation or parsing can exceed that limit. Oversize protocol input should fail the run once and terminate the child.

## Impact

A corrupted, incompatible, or custom CLI process can exhaust memory or spend excessive CPU in a process that embeds the SDK. The socket and subprocess transports expose different size guarantees for the same logical protocol.

## Suggested fix

- Use byte-oriented framing with an incremental counter rather than an unbounded JavaScript string.
- Apply the limit to both incomplete and newline-terminated lines.
- Decode UTF-8 only after a frame is within the bound.
- On overflow, stop reading, terminate and reap the child, and preserve only the existing bounded stderr tail.

## Acceptance criteria

- An exact-limit line follows the documented rule and a limit-plus-one-byte line fails.
- Overflow is detected whether the newline arrives in the same chunk or a later chunk.
- Tests cover CRLF, multibyte UTF-8 split across chunks, multiple frames per chunk, and a newline-free writer.
- Overflow settles the run once and cleans child processes and listeners.
- Socket and subprocess limits are documented together or intentionally versioned separately.

## Duplicate check

Searched open and closed issues and pull requests for SDK subprocess line limits, unbounded stdout remainders, stream-JSON framing, and oversized child output. Issues #2073 and #2074 cover different runtime/MCP and socket paths.

turkonthelurk:
Closes https://github.com/tetsuo-ai/agenc-core/issues/2092

**Draft for review only.** Not to be merged by automation; a maintainer decides when and whether it lands. Opened from fork `turkonthelurk/agenc-core`. Nothing pushed to `main` on either repo. Independent of drafts #2563 (#2090) and #2564 (#2091). Based on upstream `main` at `6a2a92d`.

## Problem

The SDK subprocess transport appended stdout into an unbounded string until it found a newline, then `JSON.parse`d each complete line. A child that emitted a very large line or never emitted a newline could consume unbounded memory and parsing time in the embedding process.

Verified still present on `tetsuo-ai/agenc-core@6a2a92d` (`main` at the time of this PR) before changing anything. No open or closed PR referenced #2092.

The socket transport already enforced a 16 MiB ceiling; the subprocess path had no equivalent.

## Change

- **`AGENC_SDK_MAX_FRAME_BYTES`** (16 MiB) is the documented shared numeric ceiling for both SDK transports, pinned to the daemon socket and MCP stdio limits.
- **Subprocess** now uses `SdkNewlineFrameDecoder`: incremental raw payload counter, UTF-8 decode only after the frame is within the bound, same rule for incomplete and newline-terminated lines. Exact 16 MiB payload is accepted; one extra byte fails whether the delimiter arrives in the same chunk or later.
- On overflow the run settles once: stop reading, SIGTERM the child, keep only the existing 8 KiB stderr tail, and do not parse or retain the oversized payload.
- **Socket is unchanged.** It still measures UTF-8 bytes of the unsliced receive buffer before line split (`MAX_CLIENT_BUFFER_BYTES` locally). A completed frame plus its delimiter can therefore trip the socket ceiling one byte earlier. The ceiling is shared; the measurement is intentionally versioned.

## Conflict risk with #2564

#2564 also edits `packages/agenc-sdk/src/subprocess.ts` (drain/`close` settlement and a wider `AgencSubprocessChild` contract) plus `docs/sdk.md` / README. This PR only replaces the stdout remainder decoder (now in `newline-frame.ts`) and adds optional `removeListener`/`pause` on stdout. The two can land in either order, but the subprocess file will need a mechanical rebase: keep #2564's exit/close settlement and this PR's `SdkNewlineFrameDecoder` / `failOversizedStdout` path. Framing tests live in new files so they do not collide with #2563/#2564 edits to `subprocess-transport.test.ts`.

#2563's subprocess touch is the prompt-event buffer; this PR leaves that drop-oldest behavior unchanged.

`socket.ts` is not in this diff, so it does not conflict with socket-side work.

## Tests

Hermetic, fail-before / pass-after:

- Decoder (64-byte ceiling): exact-limit accepted; limit-plus-one fails in the same chunk, a later chunk, or with no newline; CRLF; multibyte UTF-8 split across chunks; multiple frames per chunk.
- Subprocess wiring at the production 16 MiB bound: exact-limit then a valid result succeeds; overflow settles once, ignores a later result, drops stdout/abort listeners, preserves the stderr tail, and reaps a real hanging writer.
- Contract pin: `AGENC_SDK_MAX_FRAME_BYTES` equals the daemon/MCP 16 MiB ceilings.

```text
cd runtime && node scripts/run-hermetic-vitest.mjs run \
  tests/sdk-package/newline-frame.test.ts \
  tests/sdk-package/frame-limits.contract.test.ts \
  tests/sdk-package/subprocess-transport.test.ts \
  tests/sdk-package/subprocess-stdout-framing.test.ts
# 20 passed
npm run typecheck --workspace=@tetsuo-ai/agenc-sdk
# passed
```

This does not merge.

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/8


7etsuo:
- [newline-frame.ts:30](https://github.com/tetsuo-ai/agenc-core/blob/0b8c5b2a1a87ea911d2068e4125b76a7b1ca16c9/packages/agenc-sdk/src/newline-frame.ts#L30) counts a trailing CR, so an exact-limit payload plus CRLF is rejected, but the docs say CR is excluded; daemon.md also calls a lone CR a delimiter. Align code and docs, and test exact limit plus CRLF.
- [socket.ts:53](https://github.com/tetsuo-ai/agenc-core/blob/0b8c5b2a1a87ea911d2068e4125b76a7b1ca16c9/packages/agenc-sdk/src/socket.ts#L53) keeps its own constant, so the docs' claim that both transports share `AGENC_SDK_MAX_FRAME_BYTES` is false, and the socket can trip up to a read chunk early, not one byte. Share and test the constant, or fix the docs.
- On overflow, [subprocess.ts:236](https://github.com/tetsuo-ai/agenc-core/blob/0b8c5b2a1a87ea911d2068e4125b76a7b1ca16c9/packages/agenc-sdk/src/subprocess.ts#L236) sends one SIGTERM with no escalation. Rebase after #2563 and #2564 land (all three rewrite subprocess.ts) and use #2564's process-group kill.

PR #2570, opened 2026-09-18
fix(llm): reject provider streams that end without a terminal event (#2102)

turkonthelurk:
Closes https://github.com/tetsuo-ai/agenc-core/issues/2102

**Draft for review only.** Not to be merged by automation; a maintainer decides when and whether it lands. Opened from fork `turkonthelurk/agenc-core`. Nothing pushed to `main` on either repo. Based on upstream `main` only. Independent of #2103 / draft #2569 and drafts #2561–#2568.

## Problem

Gemini, Amazon Bedrock, and OpenAI-compatible streaming adapters treated a clean byte-stream EOF as a successful `stop` even when the provider never sent its protocol terminal event. A connection loss exactly between complete frames was therefore returned as a normal completion, including boundary-aligned partial tool calls.

Verified still present on upstream `main` at `6a2a92d` before changing anything.

## Change

Terminal state is tracked separately from the normalized finish reason. `done: true` is emitted only after that check.

| Adapter | Required terminal | Other failures |
|---|---|---|
| Gemini | Non-empty candidate `finishReason`, or recognized `promptFeedback.blockReason` | Mid-frame EOF → truncated; malformed SSE JSON → invalid |
| Bedrock | `messageStop` | Open started content/tool blocks → invalid; partial frame / bad JSON → typed errors |
| OpenAI-compatible | Choice `finish_reason` **or** `[DONE]` | `[DONE]` with open tool calls → invalid; malformed JSON → invalid |

`acceptsCleanEofAsTerminal` exists in capability gating for endpoints that document another terminal rule. No current provider enables it. Existing Z.AI / Kimi / DeepSeek strict-SSE rules are unchanged.

## Tests

Hermetic Vitest (Node 26.9.0):

- **19/19** new stream-terminal tests (EOF after a frame, EOF mid-frame, terminal + trailing usage, open blocks, cancellation, malformed JSON)
- **575** existing Gemini / Bedrock / OpenAI / Z.AI / Kimi / DeepSeek / LM Studio / Qwen / Cerebras / Groq / capability-gating tests
- **109** additional meta / OpenRouter / MiniMax / Ollama Cloud / reasoning-resume tests

**703 passed.**

This does not merge.

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/13

Issue #2095, opened 2026-09-04
fix(plugins): make install and update activation transactional


7etsuo:
## Summary

Plugin install and update replace the installed directory before writing install metadata, validating the installed copy, and updating plugin configuration. A failure in any later step leaves the command reporting failure after the old version has already been deleted and the new version has become active on disk.

Audited commit: `6d9243b2820821ac0443eaabee58fc89d156b688`

## Current behavior

[`installPluginOp`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/plugins/cli/pluginOperations.ts#L417-L499) validates the source, then calls `copyDirectoryAtomically` at lines 462-464. Only after replacement does it write metadata, validate the installed copy, and update plugin configuration at lines 465-491.

[`copyDirectoryAtomically`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/plugins/cli/pluginOperations.ts#L846-L885) stages a copy, but an update removes the existing destination at line 880 before renaming the stage at line 881. It keeps no backup and has no rollback path.

[`updatePluginOp`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/plugins/cli/pluginOperations.ts#L568-L608) delegates to the same install flow with `force: true`.

## Reproduction

1. Install version 1 of a valid plugin.
2. Prepare a valid version 2 source.
3. Inject a failure in metadata writing, installed-copy validation, or plugin-config persistence after directory replacement.
4. Run the update.

The update throws, but version 1 is gone and version 2 remains at the installed path. Depending on the failure point, install metadata or the enabled configuration is missing or stale.

A new install has the related partial-activation problem: the command can fail after leaving an installed directory that later discovery may load.

## Expected behavior

A successful return should be the commit point. Before that point, the prior installed plugin and configuration should remain usable. A failed new install should not leave a discoverable partial installation.

## Impact

Disk, permission, validation, or configuration errors can turn a routine update into an unreported partial activation. Users cannot rely on the command result to know which plugin version the next daemon start will load.

## Suggested fix

- Copy into a sibling stage and write install metadata there.
- Validate the exact staged copy, including its final identity and configuration requirements.
- Preserve the existing directory under a unique backup name before the final rename.
[...]

PR #2571, opened 2026-09-18
fix(plugins): keep cache locks with owner-checked leases (#2098)

### LINKED ISSUE #2098: fix(plugins): use owner-checked leases for cache locks
## Summary

Plugin cache locks expire after 60 seconds even though a resolution subprocess or download can run for 120 seconds. The lock has no owner token or heartbeat. Another process can remove a live lock, and the original owner's `finally` block can then remove the replacement owner's lock.

## Current behavior

Plugin process and download timeouts are 120 seconds, while the cache lock timeout is 60 seconds in [`resolution.ts` lines 144-151](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/plugins/resolution.ts#L144-L151).

[`withPluginCacheLock`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/plugins/resolution.ts#L1247-L1272) acquires a directory, runs the operation, and unconditionally removes that directory in `finally`.

[`removeStaleCacheLock`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/plugins/resolution.ts#L1274-L1283) decides ownership only from directory modification time and removes any directory older than 60 seconds.

## Reproduction

1. Let process A acquire a cache lock and keep its resolution operation running past 60 seconds.
2. Let process B attempt the same cache and remove A's lock as stale.
3. Let B acquire the replacement lock.
4. Let A finish.

A's `finally` removes B's lock because it deletes by path without checking ownership. Process C can then enter while B is still mutating the same cache.

The same ordering can be reproduced with a fake clock by aging the lock directory rather than waiting in real time.

## Expected behavior

A live owner should retain exclusive access for the full operation. Reclamation and release should delete a lock only when the caller proves that it still owns the same lease.

## Impact

Concurrent plugin installs or updates can overlap cache extraction, validation, and activation. One process can observe or replace a partial cache produced by another process.

## Suggested fix

- Store a unique owner token and process metadata in the lock.
- Refresh a lease heartbeat for the full bounded operation, or use an operating-system lock primitive.
- Reclaim only after proving lease expiry and, where possible, owner death.
- Release with compare-and-delete semantics so an old owner cannot remove a newer lease.
- Set acquisition wait independently from lease duration.

## Acceptance criteria

- An operation lasting the full 120-second process budget retains exclusive ownership.
- A second owner cannot enter while the first heartbeat is current.
- An old owner's release cannot remove a replacement owner's lock.
- A dead owner is eventually reclaimed.
- Fake-clock and multi-process tests cover A/B/C ordering, owner death, delayed heartbeat, and release races.

## Duplicate check

Searched open and closed issues and pull requests for plugin cache locks, stale-lock reclamation, lease ownership, heartbeat, and overlapping resolution. No matching report was found.

turkonthelurk:
Closes https://github.com/tetsuo-ai/agenc-core/issues/2098

**Draft for review only.** Not to be merged by automation; a maintainer decides when and whether it lands. Opened from fork `turkonthelurk/agenc-core`. Nothing pushed to `main` on either repo. Based on upstream `main` only. Independent of #2095 and drafts #2561–#2570.

## Problem

Plugin cache locks expired after 60s even though a resolve subprocess or download can run for 120s. The lock had no owner token or heartbeat, so a waiter could delete a live lock, and the original owner's `finally` could then delete the replacement lock by path.

## Change

- Extracted `withPluginCacheLock` into `runtime/src/plugins/plugin-cache-lock.ts`.
- Each lock stores a unique owner token plus pid/heartbeat metadata.
- Live owners refresh a 150s lease (independent 60s acquire wait, 30s heartbeat).
- Reclaim only after lease expiry or proven owner death (`process.kill(pid, 0)`; `EPERM` counts as alive).
- Release uses compare-and-delete (`unlink` of `owner.<token>` + `rmdir`) so an old owner cannot remove a newer lease.

This does **not** implement #2095 transactional install/update. Cache activation and install/update transaction boundaries are unchanged.

## Tests

Hermetic Vitest:

```text
node scripts/run-hermetic-vitest.mjs run \
  tests/plugins/plugin-cache-lock.test.ts \
  tests/plugins/resolution.test.ts
# 2 files, 64 passed
```

Coverage includes fake-clock / multi-owner cases for 120s exclusive ownership, current-heartbeat exclusion, delayed heartbeat reclaim, dead-owner reclaim, A/B/C release races, and a real child-process death. Restoring 60s mtime-only reclaim made the 120s exclusive-ownership test fail as expected; the lease implementation is restored and green.

This does not merge.

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/14


7etsuo:
- [plugin-cache-lock.ts:84](https://github.com/tetsuo-ai/agenc-core/blob/9c96fdd2918547f8aba687f371e61a21526c279d/runtime/src/plugins/plugin-cache-lock.ts#L84): `void lock.refresh()` turns heartbeat I/O errors into unhandled rejections that exit the daemon. Catch them, and have release await an in-flight refresh so it cannot recreate the lease.
- [plugin-cache-lock.ts:193](https://github.com/tetsuo-ai/agenc-core/blob/9c96fdd2918547f8aba687f371e61a21526c279d/runtime/src/plugins/plugin-cache-lock.ts#L193): reclaim follows a symlinked `.lock` and unlinks files in its target. `lstat` it and fail closed unless it is a real directory.
- [plugin-cache-lock.ts:223](https://github.com/tetsuo-ai/agenc-core/blob/9c96fdd2918547f8aba687f371e61a21526c279d/runtime/src/plugins/plugin-cache-lock.ts#L223): this `rmdir` can delete a lock another waiter just created, so two owners can overlap. After writing its lease, `tryAcquire` should re-check directory identity and competing leases.
- No test reaches the incomplete-dir, stale `.tmp` or corrupt-lease branches, and the fake clock (`1_000_000`) lags real mtimes. Seed it from `Date.now()` and test those branches and each fix above.

PR #2572, opened 2026-09-18
fix(plugins): make install and update activation transactional (#2095)

turkonthelurk:
Closes https://github.com/tetsuo-ai/agenc-core/issues/2095

**Draft for review only.** Not to be merged by automation; a maintainer decides when and whether it lands. Opened from fork `turkonthelurk/agenc-core`. Nothing pushed to `main` on either repo. Based on upstream `main` only. Independent of #2098 / draft #2571 (cache leases).

## Problem

Plugin install and update replaced the installed directory before writing install metadata, validating the installed copy, or updating plugin configuration. A later failure reported command failure after version 1 had already been deleted and version 2 was already live on disk. A failed first install could also leave a discoverable partial directory.

Verified still present on current `main` (`copyDirectoryAtomically` removes the destination, then renames the stage, with no backup or rollback).

## Change

Install and update now treat a successful return as the commit point:

1. Copy into a sibling stage and write install metadata there.
2. Validate that exact staged copy, including identity.
3. On update, rename the existing directory to a unique `.bak-<uuid>` before renaming the stage into place.
4. Publish plugin configuration only after the new directory is in place.
5. Remove the backup only after the new directory and configuration are durable.
6. On any pre-commit failure, restore the previous directory (update) or remove the new directory (first install).
7. Startup `loadPlugins` recovers incomplete operation records and refuses to delete or restore stage/backup objects whose identity changed.

## Tests

Hermetic Vitest — 84 tests in the focused plugin set, including 13 new rollback/crash/recovery cases:

- `tests/plugins/cli/plugin-install-transaction.test.ts` (13 new)
- plus existing plugin CLI / loader / resolution / directories / skill-metadata / source-docs suites

Acceptance covered: metadata/validation/config failures keep v1 on update and leave no discoverable first-install directory; backup survives until dest+config are durable; crash recovery at backup/replace/config boundaries; changed stage/backup objects are rejected.

This does not merge.

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/15

PR #2545, opened 2026-09-16
fix(tui): keep project tree Git state consistent

MarcoDotIO:
The project explorer now derives file badges, branch metadata, and the changed-file count from one Git status snapshot. Collapsing folders no longer hides outstanding changes, and add/add or delete/delete conflicts display as unmerged.

- Replace the two status scans per refresh with one porcelain-v2 `--branch -z` scan, publishing branch and file state together.
- Preserve rename/copy record boundaries and filenames containing whitespace or newlines; classify all seven unmerged codes.
- Use the complete snapshot count for both the WORKSPACE header and branch footer.
- Allow for v2 metadata overhead with a bounded 64 MiB status buffer. A real Git regression covers 9,000 tracked changes whose v2 output exceeds 1 MiB. Broader listing limits and failure reporting remain tracked in #2120.

Validation:
- `npm run test:fast -- --base origin/main` (passed: typechecks, 88 changed-file tests, and 734 related runtime tests).
- Focused project-tree/explorer suite (156 tests) and preview suite (20 tests); final parser/footer suite including the large-output regression (41 tests).
- `npm run build`, including package entrypoint and SDK generated-type checks, with Node 26.5.0/npm 11.17.0 and the current Xcode compiler/SDK.
- Regressions were observed failing before the relevant fixes: collapsed counts, duplicate status invocations, and larger v2 output.

Closes #2114
Closes #2115
Closes #2121

Issue #2123, opened 2026-09-04
fix(tui): check and report realtime playback capability


7etsuo:
## Summary

Realtime readiness checks only microphone recording. Playback later assumes a `play` executable exists, and spawn failures are silently reduced to player reset. A system can therefore be reported ready for voice while it cannot produce assistant audio.

Realtime start is currently disabled in production by the admission guard, so this is an activation blocker.

## Verified current behavior

- [`startDefaultRealtimeAudioCapture`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/tui/realtime/audio.ts#L40-L63) checks recording availability only.
- [`createProcessRealtimeAudioPlayer`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/tui/realtime/audio.ts#L139-L178) hardcodes the `play` command and handles spawn failure by resetting internal state without surfacing a user-visible error.
- [`checkRecordingAvailability`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/services/voice.ts#L263-L336) can return available through the native recording module or `arecord` without checking for a playback backend.
- The production RPC route remains disabled in [`realtime.ts`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/app-server/realtime.ts#L108-L113).

## Impact

When realtime is enabled, users can enter a session that records and sends audio but silently discards all returned speech.

## Fix direction

Probe a supported playback backend as part of realtime capability detection. Surface startup and runtime playback failures through controller state. If native audio supports output, prefer it or document and verify the external dependency per platform.

## Acceptance criteria

- Realtime readiness covers both recording and playback.
- Missing or failed playback reports an actionable error.
- Capability checks cover macOS, Linux, Windows, and remote environments.
- Tests cover a working microphone with no playback command.
- The check is in place before production realtime admission is enabled.

## Duplicate check

GitHub issue and pull-request searches for realtime playback availability, missing SoX `play`, and voice readiness found no matching report.

Issue #2122, opened 2026-09-04
fix(tui): honor realtime audio backpressure before draining queued chunks


7etsuo:
## Summary

The realtime audio player records that child stdin is waiting for `drain`, but every subsequent enqueue calls `flush` again. `flush` does not check that state, so it continues writing before the stream drains and bypasses the managed 512 KiB queue limit.

Realtime start is currently disabled in production by the admission guard, so this is an activation blocker.

## Verified current behavior

[`flush`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/tui/realtime/audio.ts#L94-L124) sets `waitingForDrain` after `stdin.write` returns false but does not return early when called again in that state. [`enqueueBuffer`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/tui/realtime/audio.ts#L126-L136) calls `flush` after every chunk.

With a fake stdin that always returns false, enqueueing twenty 64 KiB chunks caused 1,310,720 bytes to be written before any `drain` event. The queue constant is 512 KiB.

The production RPC route is disabled in [`realtime.ts`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/app-server/realtime.ts#L108-L113).

## Impact

When realtime is enabled, slow playback can move unbounded data into Node and OS stream buffers, increasing memory use and playback latency while defeating the intended drop policy.

## Fix direction

Make `flush` return immediately while `waitingForDrain` is true. Resume only from the registered `drain` handler. Define and test whether the bounded queue drops oldest or newest audio.

## Acceptance criteria

- No additional `stdin.write` occurs between a false return and `drain`.
- Buffered audio under application control never exceeds the documented cap.
- Error, close, and format-change paths clear the drain listener and queue safely.
- The fix is covered before production realtime admission is enabled.

## Duplicate check

GitHub issue and pull-request searches for realtime audio drain handling, playback backpressure, and the 512 KiB queue found no matching report.

PR #2573, opened 2026-09-18
fix(tui): generation-fence realtime capture startup and callbacks (#2113)

### LINKED ISSUE #2113: fix(tui): generation-fence realtime capture startup and callbacks
## Summary

Realtime websocket audio capture startup and callbacks are not tied to a controller generation. A capture promise can resolve after the session has stopped or restarted, install an obsolete capture as current, and send late audio or terminal callbacks into the new state.

Realtime start is currently disabled in production by the admission guard, so this is an activation blocker rather than a currently reachable production path.

## Verified current behavior

[`#startWebsocketAudioCapture`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/tui/realtime/controller.ts#L451-L475) awaits `#startAudioCapture` and assigns its result without checking that the initiating session is still current. The supplied `onAudio`, `onLevel`, `onError`, and `onClosed` callbacks also have no generation check.

The production RPC route is explicitly disabled in [`realtime.ts`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/app-server/realtime.ts#L103-L113).

## Impact

When realtime admission is enabled, a slow capture backend can resurrect capture after stop, leak input into another session generation, or tear down the wrong active session through a late callback.

## Fix direction

Allocate a generation token before capture startup. After the promise resolves, stop and discard the capture unless the token is still current. Apply the same check to every callback.

## Acceptance criteria

- A capture resolving after stop is immediately stopped and never becomes current.
- Callbacks from an older generation cannot append audio, update levels, or close a newer session.
- Tests cover stop-during-start and restart-during-start races.
- The guard is in place before production realtime admission is enabled.

## Duplicate check

GitHub issue and pull-request searches for realtime capture races, stale callbacks, and session generations found no matching report.

turkonthelurk:
Closes https://github.com/tetsuo-ai/agenc-core/issues/2113

**Draft for review only.** Not to be merged by automation; a maintainer decides when and whether it lands. Opened from fork `turkonthelurk/agenc-core`. Nothing pushed to `main` on either repo. Based on upstream `main` only. Independent of drafts #2561–#2572 and TUI Git PR #2545. Does **not** implement #2122 or #2123. Does **not** enable production realtime admission.

## Problem

Websocket audio capture startup and callbacks were not tied to a controller generation. A capture promise could resolve after the session stopped or restarted, install an obsolete capture as current, and send late audio or terminal callbacks into the new state.

Realtime start remains disabled in production by the admission guard; this is an activation blocker.

## Change

- Increment a capture generation when stopping capture (including during an in-flight start).
- After `#startAudioCapture` resolves, stop and discard the session unless that generation is still current.
- Guard every capture callback (`onAudio`, `onLevel`, `onError`, `onClosed`), including the async `appendAudio` failure path, so stale callbacks cannot affect a newer session.

## Tests

Hermetic Vitest (Node 26.9.0): **12 files, 79 passed**, including:

- stop-during-start discard
- discard-stop failure logging
- restart-during-start stale callbacks
- stale in-flight append failure
- `stop_requested` terminal callbacks
- existing controller / audio / admission-guard contracts

This does not merge and does not enable production realtime admission.

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/16


7etsuo:
- [controller.ts:458](https://github.com/tetsuo-ai/agenc-core/blob/6f3233c14bebe0c2d76dd0749cdcc33766545e7c/runtime/src/tui/realtime/controller.ts#L458) reads the generation only after the `thread/realtime/start` RPC. A `realtime_closed` or `realtime_error` handled during that RPC therefore goes unfenced, and start still installs a live mic capture in an inactive session that `stop()` cannot clear. Return before `#startAudioCapture` when `!this.#canApplyRealtimeSessionEvent()`, and add a test whose start handler closes the session before resolving.
- [controller.ts:488](https://github.com/tetsuo-ai/agenc-core/blob/6f3233c14bebe0c2d76dd0749cdcc33766545e7c/runtime/src/tui/realtime/controller.ts#L488): `Promise.resolve(capture.stop())` does not catch a synchronous throw, and the default capture stops synchronously, so `start()` rejects for a session that has already closed. Use `Promise.resolve().then(() => capture.stop())` and cover it with a test.

PR #2574, opened 2026-09-18
fix(tui): honor realtime audio backpressure before draining queued chunks (#2122)

### LINKED ISSUE #2122: fix(tui): honor realtime audio backpressure before draining queued chunks

turkonthelurk:
Closes https://github.com/tetsuo-ai/agenc-core/issues/2122

**Draft for review only.** Not to be merged by automation; a maintainer decides when and whether it lands. Opened from fork `turkonthelurk/agenc-core`. Nothing pushed to `main` on either repo. Based on upstream `main` only. Independent of #2113 / draft #2573 and #2123. Does **not** enable production realtime admission.

## Problem

`createProcessRealtimeAudioPlayer` recorded stdin backpressure (`waitingForDrain`) but `flush()` did not return while that flag was set. Every later `enqueueBuffer()` called `flush()` again, so `stdin.write` kept going after a `false` return and bypassed the managed 512 KiB queue.

Verified on upstream `main` at `6a2a92d`: a fake stdin that always returns `false` accepted all twenty 64 KiB chunks (1,310,720 bytes) before any `drain`.

## Change

- `flush()` returns immediately while `waitingForDrain` is true.
- Further writes resume only from the registered `drain` handler (and ignore a stale child after reset).
- Error / close / format-change paths detach the drain listener and drop the queue.
- Drop policy: **oldest queued audio is dropped** so the newest frames stay and the application-owned queue never exceeds 512 KiB.

## Tests

Hermetic Vitest, fail-before / pass-after:

- RED: 20 writes / 1,310,720 bytes before `drain`
- GREEN: 1 write / 65,536 bytes before `drain`; after `drain`, only the newest 512 KiB is flushed
- `tests/tui/realtime`: **7 files, 57 passed**

This does not merge and does not enable production realtime admission.

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/17


7etsuo:
The close test at [audio.contract.test.ts:207](https://github.com/tetsuo-ai/agenc-core/blob/ed32de50c13bd3077e18b7097768772bac7ef66d/runtime/tests/tui/realtime/audio.contract.test.ts#L207) fails once merged with main. Since #2657, `close()` sends SIGTERM only when `isSignalablePid(active.pid)` passes, and the fake from [createChild()](https://github.com/tetsuo-ai/agenc-core/blob/ed32de50c13bd3077e18b7097768772bac7ef66d/runtime/tests/tui/realtime/audio.contract.test.ts#L31) has no pid, so `kill` is never called. Give it one with `Object.assign(child, { pid: 424_242 })`, as `swarm-106-realtime-audio.test.ts` does. Then rebase onto main and post a green run of `tests/tui/realtime` and `swarm-106-realtime-audio.test.ts`.

PR #2577, opened 2026-09-18
feat(transcript-v2): report plan-mode state from the pass that builds the snapshot

signerless:
## Why

Opening a session with a long history hung the desktop renderer. Measured from
the running app: `transcript()` on a 10-entry session with 1857 turn-events took
45.4 s, 45.3 s and 77.0 s on three consecutive calls, a 6-entry session 217 ms,
a brand-new empty session 3.5 s. The desktop bridge pages the entire run journal
through `run.replay` (200 per page, 112 RPCs for `asOfSequence` 22,295) on every
open, solely to find the latest `run_runtime_settings_changed` event and derive
`planModeActive`. Core already walks that same history to build the transcript.

## What, per file

- `runtime/src/app-server/protocol/index.ts` — `SessionTranscriptV2Result` gains
  optional `planModeActive` and `planModeSequence`, documented as the plan-mode
  state at `asOfSequence` from the same history the transcript was rebuilt from.
- `runtime/src/app-server/background-agent-runner/journal-reconstruction.ts` —
  the existing `asOfSequence` loop in `sessionTranscriptV2FromRollout` also keeps
  the latest runtime-settings event by sequence; the result includes the pair
  only when such an event exists. No extra I/O, no second pass.
- `packages/agenc-sdk/src/transcript-v2.generated.ts`,
  `packages/agenc-sdk/src/protocol-wire.generated.ts` — regenerated with
  `check-sdk-generated-types --write`.
- `runtime/tests/app-server/transcript-v2.contract.test.ts` — typed
  `settingsEvent` fixture and a case covering: a later "default" overriding an
  earlier "plan" chosen by sequence not array order; plan-only history; empty
  history reporting nothing.

## Evidence

Root cause confirmed from source and measurement in the desktop bridge:
`resumeSession` is 0 ms; the replay loop is the cost; the slow session's rollout
is 5.66 MB / 22,367 lines versus 126 KB for the fast one. Stored event shape
verified from a real plan-mode session rollout: `event_msg` /
`msg.type = run_runtime_settings_changed` / `msg.payload.permissionMode`.

## Tests

`run-hermetic-vitest` on `transcript-v2.contract.test.ts` and
`daemon-transcript-snapshot-notices.test.ts`: all pass. `tsc --noEmit`: zero
non-baseline errors (TS2883 baseline unchanged). `check-sdk-generated-types
--check`: verified committed SDK types.

## Not changed

`run.replay` itself, `schema.json` (method-level only; result types are not
declared there), every other field of the V2 result, and all consumers.

## Counterpart PRs

agenc-desktop #293 uses the pair when present and keeps the journal replay as
the fallback for daemons that predate this change. Either can merge first.


signerless:
Operational measurement on the campaign machine after relaunching the desktop dev build on this core (`d39e1f9bc`) with the desktop counterpart (`ec2f853a2`), same `AGENC_HOME`, sessions recovered after the daemon restart and resumed before opening (what the UI does):

```
resume=   265ms  transcript=    43ms  entries=6   planModeSequence=3  task05 small
resume=     7ms  transcript=     6ms  entries=9   planModeSequence=3  task02 medium
resume=    44ms  transcript=    69ms  entries=10  planModeSequence=4  task01 LARGE (45-77 s before the fix)
resume=     -    transcript=    65ms  entries=10  planModeSequence=4  task01 second open
resume=     -    transcript=    40ms  entries=6   planModeSequence=3  task05 second open
```

The large session (`conv-mu6ltuyj`, ~22k journal events) opened in 45-77 s before; it now opens in 69 ms with `planModeSequence` reported by the daemon, so the run-journal replay fallback was not taken.

PR #2582, opened 2026-09-18
fix(session): scope a mid-turn provider switch to the turn in flight

signerless:
## Why

Switching model or provider while a turn is running left the session dead: the current turn aborted as designed, but every later prompt was admitted, reached `running` and settled with `provider_switched` a few seconds later without any assistant output or provider request. Reproduced on the desktop campaign daemon (session `conv-mu6niob2`, turns 334/338/342) and independently by Codex with a cold daemon and a loopback provider (turns `sub-conv-mu6odtlj-19/23`).

Cause: `commands/model.ts:239-252` and `commands/provider.ts:174-176` call `Session.abortTerminal("provider_switched")`. That method aborts the session's readonly lifetime `AbortController` (`session.ts:2540`), and every `run-turn` merges that signal with the task's own controller (`run-turn.ts:2483-2489`). A lifetime token cannot be un-aborted, so the staged switch never got a turn to apply to.

## What

- `runtime/src/session/session.ts`: `abortTerminal` routes `provider_switched` to a new private `abortActiveTurnForProviderSwitch()`, which aborts the `ActiveTurn` controller and each task controller of the turn in flight with the same reason and emits `turn_aborted` for that turn id. Any other reason keeps the previous terminal behaviour. Nothing touches `pendingProviderSwitch`; the next prompt consumes it at top-of-loop (`consumeDaemonPendingProviderSwitches`).
- `runtime/tests/session/session.test.ts`: two cases in `Session.abortTerminal`: the turn and child controllers carry `provider_switched`, the session controller stays clean and a later `stdin_lost` still terminates; `provider_switched` with no active turn is a no-op.

## Evidence

- Live desktop: after one mid-turn switch meta -> deepseek, three follow-up prompts settled with `stopReason stopped / exit 130 / finalMessage provider_switched` in ~2 s each; an idle `setSessionModel` to the same target answered "Model unchanged" and did not recover the session.
- Codex cold repro: `MODEL_SWITCH_0808_REVIEW.md` (loopback provider, no request reaches the new provider for the follow-up turns).

## Tests

- `npx vitest run tests/session/session.test.ts tests/commands/model.test.ts`: 141 passed.
- `tsc --noEmit`: only the pre-existing TS2883 lodash `MemoizedFunction` baseline.
- End-to-end on a build with this fix (desktop driver: idle switch, mid-turn switch, two follow-ups completing on the new provider) is queued in the campaign and will be reported on this PR.

## Not changed

- `commands/model.ts` / `commands/provider.ts` still call `abortTerminal("provider_switched")`; the TUI and bridge shims keep their optional `abortTerminal` contract.
- The SDK reports `stopReason: completed, exitCode: 0` for these canonical aborts while the desktop reports `stopped / 130` (Codex finding). Separate outcome-mapping follow-up.
- Duplicate `turn_aborted` events for one turn (seq 31 and 32 in the Codex trace) are not addressed here.

## Counterpart PRs

None. Desktop needs no change.


signerless:
Live end-to-end on a build containing this PR (integration dist `c515eed8d`, desktop dev app, session `conv-mu6p5xne`):

```
A  deepseek-flash turn                          completed  10.1 s
   idle switch -> meta/muse-spark-1.3           applied
B  muse-spark-1.3 turn                          completed  20.8 s
   mid-turn switch -> deepseek/deepseek-flash   "staged ... current turn aborted" (turn 269 aborted, provider_switched)
   follow-up 1                                  completed  9.0 s  turnResults.provider=deepseek model=deepseek-flash
   follow-up 2 ("name your provider and model") completed  3.5 s  reply: "Provider: deepseek; model: deepseek-flash."
```

Before the fix the same driver saw every follow-up settle with `provider_switched` in ~2 s and no assistant output (turns 334/338/342 on `conv-mu6niob2`). Follow-up 1 even finished the work the aborted turn had started.

PR #2583, opened 2026-09-18
fix(routines): hand routine agents the daemon's client environment

signerless:
## Why

Every routine run on the desktop campaign daemon failed in 1-5 s with `sessionId: null` and "Routine could not run. Check its workspace, provider configuration, and session details.", including the hourly "Code briefing" template routine (failed at 02:00, 03:00 ... 07:00Z on 2026-09-18) that completed on 2026-09-09. Workspace identity was intact. A cold isolated repro with the swallowed cause printed shows the real error:

```
cerebras provider requires credentials. Set CEREBRAS_API_KEY.
deepseek provider requires credentials. Set DEEPSEEK_API_KEY.
```

with both keys present in the daemon process environment. Daemon-created agents inherit nothing from that environment by design (`agent-lifecycle.ts:970-974`: the daemon materializes exactly the client's `envOverrides`); the desktop forwards its stored keys per session, the routine executor forwarded nothing. Two defects: routines cannot use any keyed provider (H-6), and `routines/service.ts` discards the cause without logging it (H-5).

## What

- `runtime/src/routines/daemon-executor.ts`: `providerEnvironmentKeys(provider)` reads the built-in provider registry (credential fields and base-URL override of that provider only); `routineSessionEnvironment(env, provider)` forwards PATH plus those variables from the daemon process, dropping empty values. `createDaemonRoutineExecutor` takes `environment` (snapshotted at construction) and `defaultProvider()` (used when the routine names no provider), and passes `envOverrides` to `createAgent` when non-empty. Nothing else crosses: no `AGENC_CREDENTIAL_*` bearers, no session or remote tokens, no browser flags, no other provider's key.
- `runtime/src/routines/service.ts`: optional `onRunFailure` sink receives `{ routineId, runId, reason, errorName?, errorCode? }` with `reason` in `credentials_missing | workspace_changed | unknown`, the error class name and an identifier-shaped `code` only; cause text never reaches the sink. `run.error` is a fixed classified string (provider name for missing credentials, changed workspace, otherwise the previous generic text).
- `runtime/src/app-server/daemon-cli.ts`: passes `environment: host.env` and `defaultProvider: () => resolveBuiltInProviderSlug(activeConfig.model_provider)` (live config, reload-safe) to the executor; logs `agenc: routine <id> run <id> could not run: <reason> [CODE] (ErrorName)`.
- `runtime/tests/routines/daemon-executor.test.ts`: forwards only PATH and the provider's own variables; default provider path; unknown provider and nothing-forwardable cases; negative test for MCP bearers, session/remote tokens, browser flags, `AGENC_PROVIDER`/`AGENC_MODEL` and other providers' keys; registry-derived keys.
- `runtime/tests/routines/service.test.ts`: credentials cause -> sink reason `credentials_missing`, provider named in `run.error`, env var name absent from both; unknown secret-bearing cause -> reason `unknown` with class name and code, secret absent from sink and record, non-identifier code dropped; changed workspace; throwing sink.

## Evidence

- Live store `routines-v1.json`: 6 consecutive hourly failures with `sessionId: null`; three manual routines created during the campaign (fresh cwd, trusted cwd, cerebras with the key in the daemon env) all failed identically.
- Cold repro script (`probe.mjs` + chunk patch in a restored copy): both providers fail in 60 ms at `createAgent` with the credentials error above; daemon env keys listed by name in the run log.

## Tests

- `npx vitest run tests/routines tests/app-server/daemon-dispatcher.routines.contract.test.ts`: 53 passed (4 files).
- `tsc --noEmit`: only the pre-existing TS2883 lodash baseline.
- Cold end-to-end on the first revision: routines complete on deepseek (7.0 s) and cerebras (4.9 s) with real sessions (comment below); rerun on this revision follows.
- Desktop-driven routine on the integration build: session created, four files read, `Write` refused by the unattended policy with no approver attached (expected for default mode; product follow-up H-9).

## Review

Second revision addresses both Codex findings: forwarding is now scoped to the selected provider (was the whole canonical client environment), and the diagnostic carries a fixed reason instead of the raw cause.

## Not changed

- Keys the desktop stores after the daemon started are still invisible to routines until the daemon restarts; the desktop forwards those per session today, a routine-side equivalent is a desktop follow-up.
- `agent.create` credential resolution and the `AGENC_DAEMON_CLIENT_ENV_KEYS` allowlist are untouched.
- Scheduled runs that already failed are not retried.

## Counterpart PRs

None required. Desktop follow-up noted above.


signerless:
Cold end-to-end on a build containing this PR (integration dist `c515eed8d`, isolated `AGENC_HOME`, keys only in the daemon process env, same probe that reproduced the failure):

```
deepseek/deepseek-flash   status=completed  sessionId=session_b6db2cce-…  7.0 s
cerebras/qwen-3.8-27b     status=completed  sessionId=session_f16d90e9-…  4.9 s
```

Before the fix the same probe failed both in 60 ms with `sessionId: null` and the cause `<provider> provider requires credentials. Set <PROVIDER>_API_KEY.` printed from a patched copy of the chunk. Desktop-driven routine run on the relaunched app is running now and will be added here.

signerless:
Second revision (`24ede9ab9`, provider-scoped forwarding + structured diagnostic) rebuilt into the integration dist `6f2b0ccb9` and cold-probed again (isolated home, keys only in the daemon process env, routine created and run over the public RPC):

```
deepseek/deepseek-flash   status=completed  sessionId=session_de61e122-…  7.5 s
cerebras/qwen-3.8-27b     status=completed  sessionId=session_f5046668-…  8.8 s
```

Same probe as before; the forwarded environment for each run is now exactly `PATH` plus that provider's own credential and base-URL variables.

Issue #2106, opened 2026-09-04
fix(llm): map Gemini prompt-level blocks instead of returning empty success


7etsuo:
## Summary

Gemini responses with a prompt-level block and no candidates are parsed as successful empty completions. The adapter ignores `promptFeedback.blockReason`, and its missing finish-reason fallback is `stop`.

Audited commit: `6d9243b2820821ac0443eaabee58fc89d156b688`

## Current behavior

[`readCandidateParts`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/gemini/index.ts#L2620-L2632) returns an empty part list when the response has no candidates. [`parseGeminiResponse`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/gemini/index.ts#L2644-L2665) does not inspect top-level prompt feedback. [`geminiFinishReason`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/gemini/index.ts#L162-L175) maps an absent reason to stop.

Google documents that a set `promptFeedback.blockReason` means the prompt was blocked and no candidates are returned: [GenerateContentResponse](https://ai.google.dev/api/generate-content) and [safety settings](https://ai.google.dev/gemini-api/docs/safety-settings).

## Reproduction

Parse this valid provider response in chat and streaming paths:

```json
{
  "promptFeedback": {
    "blockReason": "SAFETY"
  },
  "usageMetadata": {
    "promptTokenCount": 3,
    "totalTokenCount": 3
  }
}
```

The adapter returns empty content with `finishReason: "stop"` instead of reporting a filtered prompt.

## Expected behavior

The adapter should evaluate prompt-level feedback before candidate parsing. Recognized content-policy blocks should map to `content_filter`; unknown or operational block reasons should become typed provider errors with useful diagnostics.

## Impact

Callers cannot distinguish a blocked request from a legitimate empty answer. Session logic can persist the turn as successful and continue without explaining why no output exists.

## Suggested fix

- Parse `promptFeedback.blockReason` and `safetyRatings` before reading candidates.
- Map safety, blocklist, prohibited-content, and image-safety reasons to the content-filter outcome.
- Treat unexpected reasons as provider errors rather than normal stop.
[...]

Issue #2105, opened 2026-09-04
fix(llm): preserve Ollama done_reason truncation


7etsuo:
## Summary

The Ollama adapter ignores `done_reason` in both streaming and non-streaming responses. It maps every terminal response without tool calls to a normal stop, including `done_reason: "length"`.

Audited commit: `6d9243b2820821ac0443eaabee58fc89d156b688`

## Current behavior

The streaming path hardcodes finish reason from tool-call presence at [`ollama/adapter.ts` lines 678-680](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/ollama/adapter.ts#L678-L680).

The non-streaming parser does the same at [`ollama/adapter.ts` lines 995-1022](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/ollama/adapter.ts#L995-L1022).

Ollama's chat API includes `done_reason` in the terminal response: [chat API](https://docs.ollama.com/api/chat). Ollama itself emits `length` when generation reaches a limit and maps that value to truncation in its compatibility layers.

## Reproduction

Return a valid terminal Ollama response with content, `done: true`, and `done_reason: "length"`. Test both stream and non-stream modes.

Both modes return `finishReason: "stop"`.

## Expected behavior

Ollama terminal reasons should be normalized into AgenC's finish-reason contract. `length` must become `length`, and unknown non-empty reasons should produce a diagnostic or explicit error rather than silently becoming a natural stop.

## Impact

Truncated prose and tool JSON can be accepted as complete. AgenC's maximum-output warning, continuation logic, and tool suppression for truncated output never run.

## Suggested fix

- Capture `done_reason` from the final streaming chunk and the non-streaming response.
- Map known values through one Ollama-specific normalizer.
- Preserve the raw reason in diagnostics when it is unknown.
- Do not return incomplete tool calls when the reason indicates truncation.

## Acceptance criteria

- `length` maps to AgenC `length` in both modes.
- Natural `stop` retains current successful behavior.
- Tool-call and truncation precedence is defined and tested.
- Missing and unknown reasons produce a documented fallback with diagnostics.
- Streaming uses the terminal chunk's reason even when earlier chunks omit it.

## Duplicate check

[...]

Issue #2104, opened 2026-09-04
fix(llm): include Gemini thinking tokens in output accounting


7etsuo:
## Summary

Gemini usage normalization stores `candidatesTokenCount` as total completion output and `thoughtsTokenCount` as a separate reasoning field. AgenC's cost and budget contracts treat reasoning tokens as a subset of output tokens, so Gemini thinking tokens are omitted from output cost and output-budget accounting.

Audited commit: `6d9243b2820821ac0443eaabee58fc89d156b688`

## Current behavior

[`requestUsageFromGemini`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/gemini/index.ts#L151-L159) maps:

- `completionTokens` to `candidatesTokenCount`
- `reasoningOutputTokens` to `thoughtsTokenCount`

The cost contract says reasoning is a subset of output in [`session/cost.ts` lines 540-560](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/session/cost.ts#L540-L560). The session accumulator adds only `completionTokens` to `outputTokens` at [`cost.ts` lines 1420-1434](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/session/cost.ts#L1420-L1434). Admission cost and turn-boundary budget checks also use completion output at [`admitted-model-call.ts` lines 243-268](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/budget/admitted-model-call.ts#L243-L268) and [`stream-model.ts` lines 1543-1551](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/phases/stream-model.ts#L1543-L1551).

Google reports candidate and thinking counts separately and says response pricing is the sum of output tokens and thinking tokens: [Gemini thinking pricing](https://ai.google.dev/gemini-api/docs/generate-content/thinking) and [token counting](https://ai.google.dev/gemini-api/docs/generate-content/tokens).

## Reproduction

Normalize usage containing:

```json
{
  "promptTokenCount": 4,
  "candidatesTokenCount": 2,
  "thoughtsTokenCount": 100,
  "totalTokenCount": 106
}
```

[...]

PR #2588, opened 2026-09-18
fix(llm): prevent Anthropic fallback retries after thinking is emitted (#2107)

### LINKED ISSUE #2107: fix(llm): prevent Anthropic fallback retries after thinking is emitted
## Summary

Anthropic streaming fallback retries are allowed after thinking blocks have already been sent to the consumer. The retry guard checks emitted text and tool calls but not emitted thinking events, so a transparent retry can replay the same thinking stream and leave unbalanced or duplicated UI state.

## Current behavior

The adapter emits thinking block starts, deltas, and stops at [`anthropic/adapter.ts` lines 580-640](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/anthropic/adapter.ts#L580-L640) and [lines 663-709](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/anthropic/adapter.ts#L663-L709).

The in-stream error retry check at [`anthropic/adapter.ts` lines 745-775](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/anthropic/adapter.ts#L745-L775) tests only content, completed tools, and open tool blocks.

The outer retry check at [`anthropic/adapter.ts` lines 856-887](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/anthropic/adapter.ts#L856-L887) uses the same incomplete definition of emitted output. Neither includes open or completed thinking blocks.

## Reproduction

1. Have the first stream emit a thinking block start and one thinking delta.
2. Return a retryable provider error before any text or tool call.
3. Let the configured fallback retry succeed and emit its thinking plus final text.

The consumer sees thinking events from both attempts. The returned final response contains only the successful attempt, so rendered and persisted state disagree. If the first attempt failed before its block stop, the event sequence is also unbalanced.

## Expected behavior

Once any model-visible event has been delivered to the consumer, the adapter should not replay the request transparently unless it first emits a typed reset that every consumer understands.

## Impact

The TUI and SDK can display duplicate reasoning or leave a thinking block open while the canonical response contains a different single attempt. Streaming token estimates can also count both attempts.

## Suggested fix

- Include thinking starts, deltas, and completed blocks in the emitted-output retry guard.
- Prefer returning a partial provider error after the first emitted event.
- If transparent reset semantics are desired, define one public generation-reset event and update every consumer before using it.

## Acceptance criteria

- A retryable error after a thinking start or delta does not silently restart the stream.
- Thinking block starts and stops remain balanced on every failure path.
- The rendered event history and final persisted response describe the same attempt.
- Tests cover errors after start, delta, signature, stop, text, and tool events.
- Pre-output retry behavior remains unchanged.

## Duplicate check

Searched open and closed issues and pull requests for Anthropic fallback retries, duplicated thinking, unbalanced thinking blocks, and emitted-output guards. PR #2015 changed the outer retry closure but did not include thinking events.

turkonthelurk:
Closes https://github.com/tetsuo-ai/agenc-core/issues/2107

**Draft for review only.** Not to be merged by automation; a maintainer decides when and whether it lands. Opened from fork `turkonthelurk/agenc-core`. Nothing pushed to `main` on either repo. Based on upstream `main` only. Independent of #2104/#2586, #2105/#2587, and #2106.

## Problem

Anthropic streaming fallback retries were allowed after thinking blocks had already been forwarded. The emitted-output guard checked text and tool calls only, so a retryable error after a thinking start or delta could replay the thinking stream and leave unbalanced or duplicated UI state.

## Change

- Both retry guards (in-stream `error` and outer catch) treat open and completed thinking blocks as emitted output, same as text and tool calls.
- After a thinking start/delta/signature/stop has been forwarded, a retryable error fails closed: no silent restart, open thinking blocks are closed so start/stop stay balanced, and the adapter returns a partial error whose `thinking` matches the streamed attempt.
- Pre-output fallback retry is unchanged.
- No generation-reset event was added.

## Tests

Hermetic Anthropic adapter suite: **70 passed / 7 files**, including 13 new #2107 cases (in-stream and transport overload after start, delta, signature, stop, text, and tool; plus unchanged pre-output retry).

This does not merge.

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/27


7etsuo:
The `thinking !== undefined ||` short-circuit in [adapter.ts:1015](https://github.com/tetsuo-ai/agenc-core/blob/24f295a360f303165768b0fedc88e22bd11f153d/runtime/src/llm/providers/anthropic/adapter.ts#L1015) returns a partial whenever thinking was emitted. Since thinking streams first, mid-stream ECONNRESET, `terminated`, and 5xx faults skip the #2463 reconnect ladder, and a thinking-only abort skips the `stream_idle` conversion. Return a partial only when the fault can't be re-sampled, and rethrow on abort. The comment claiming invalid tool_use JSON still rethrows is false after thinking; fix one or the other. The [transport cases](https://github.com/tetsuo-ai/agenc-core/blob/24f295a360f303165768b0fedc88e22bd11f153d/runtime/tests/llm/providers/anthropic/adapter.thinking-stream-retry.test.ts#L181) only use non-transient `overloaded_error`; add tests for transient faults, thinking plus a streamed tool call, an abort, and a protocol error after thinking.

Issue #2112, opened 2026-09-04
fix(llm): parse Anthropic thinking-token usage details


7etsuo:
## Summary

Anthropic now reports its reasoning breakdown at `usage.output_tokens_details.thinking_tokens`. AgenC reads only a flat `reasoning_output_tokens` field, so Anthropic thinking usage is absent from telemetry in both chat and streaming paths.

Audited commit: `6d9243b2820821ac0443eaabee58fc89d156b688`

## Current behavior

Non-stream parsing reads `usageRecord.reasoning_output_tokens` at [`messages-anthropic.ts` lines 514-525](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/wire/messages-anthropic.ts#L514-L525).

The streaming accumulator type and merge logic also read only `reasoning_output_tokens` at [`anthropic/adapter.ts` lines 70-75](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/anthropic/adapter.ts#L70-L75) and [lines 99-148](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/anthropic/adapter.ts#L99-L148).

Anthropic's current Messages schema exposes `usage.output_tokens_details.thinking_tokens`. It states that `output_tokens` remains the inclusive billed total and the nested field is its reasoning subset: [Messages API types](https://platform.claude.com/docs/en/api/typescript/messages).

## Reproduction

Parse a response with this usage object:

```json
{
  "input_tokens": 120,
  "output_tokens": 348,
  "output_tokens_details": {
    "thinking_tokens": 312
  }
}
```

AgenC returns completion 348 but no `reasoningOutputTokens`. The same omission occurs when the field arrives in streaming usage events.

## Expected behavior

`completionTokens` should remain the inclusive 348. `reasoningOutputTokens` should be 312, giving observability without adding the reasoning count to the billed total again.

## Impact

Usage displays, diagnostics, and any separate reasoning-cost breakdown report zero Anthropic thinking even when most output tokens were reasoning. Total output billing remains correct because `output_tokens` is already inclusive.

## Suggested fix

- Parse `output_tokens_details.thinking_tokens` in non-stream and stream usage.
[...]

Issue #2110, opened 2026-09-04
fix(llm): preserve pre-aborted signals in local provider health checks


7etsuo:
## Summary

The shared local-provider health wrapper adds an abort listener but does not check whether the caller's signal was already aborted. Ollama and LM Studio operations can begin and complete after cancellation when the signal was aborted before entry.

Audited commit: `6d9243b2820821ac0443eaabee58fc89d156b688`

## Current behavior

[`runLocalProviderHealthSidecar`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/shared/local-health.ts#L31-L59) creates a new controller and registers `abortFromUpstream` with `addEventListener`.

`AbortSignal` does not replay an earlier abort event to a later listener. The wrapper does not call `throwIfAborted()` or invoke `abortFromUpstream` when `params.signal.aborted` is already true.

## Reproduction

1. Create an `AbortController` and abort it with a known reason.
2. Call `runLocalProviderHealthSidecar` with that signal.
3. Have the operation record its received signal state and resolve normally.

The operation receives `aborted: false` and the wrapper returns success.

## Expected behavior

A pre-aborted caller signal should prevent the operation and health timer from starting, or the operation should receive an already-aborted derived signal with the exact caller reason.

## Impact

Cancelled Ollama and LM Studio turns can begin provider work. Combined with a provider path that does not physically abort, this can hold resources after the user has cancelled.

## Suggested fix

- Check `params.signal?.aborted` before creating timers or invoking the operation.
- Abort the derived controller with `params.signal.reason` and reject through the same cancellation contract used for in-flight aborts.
- Keep current listener cleanup for signals that abort later.

## Acceptance criteria

- A pre-aborted signal prevents provider work and preserves its reason.
- An abort between setup and operation start is not lost.
- In-flight abort behavior and consecutive health-failure behavior remain unchanged.
- No timer or listener remains after early rejection.
- Tests cover pre-abort, setup race, in-flight abort, and normal completion.

## Duplicate check

Searched open and closed issues and pull requests for local-provider health checks, pre-aborted signals, Ollama cancellation, and LM Studio cancellation. No matching report was found.

PR #2592, opened 2026-09-18
fix(llm): enforce or reject Ollama toolChoice constraints (#2109)

### LINKED ISSUE #2109: fix(llm): enforce or reject Ollama toolChoice constraints
## Summary

The Ollama adapter never reads `options.toolChoice`. Calls using `none`, `required`, or a specific tool produce the same request body and can violate the caller's tool-use constraint.

## Current behavior

[`OllamaAdapter.buildParams`](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/ollama/adapter.ts#L799-L850) attaches the selected tool catalog but never checks `options.toolChoice`.

Request diagnostics hardcode `toolChoice: undefined` at [`ollama/adapter.ts` lines 141-175](https://github.com/tetsuo-ai/agenc-core/blob/6d9243b2820821ac0443eaabee58fc89d156b688/runtime/src/llm/providers/ollama/adapter.ts#L141-L175), so traces also hide the requested constraint.

Ollama's native chat API accepts a tools list but does not expose OpenAI-style tool-choice enforcement: [Ollama chat API](https://docs.ollama.com/api/chat). AgenC still needs to honor constraints it can enforce and reject those it cannot.

## Reproduction

Build requests with the same message and tools but these choices:

- `toolChoice: "none"`
- `toolChoice: "required"`
- a specific tool choice

All request bodies contain the same full tool catalog. No local validation or post-response enforcement distinguishes the choices.

## Expected behavior

`none` should omit tools so the provider cannot call them. Required or specific choices should be enforced through a supported Ollama mechanism, constrained catalog plus response validation where sound, or rejected locally as unsupported.

## Impact

A call that explicitly forbids tools can still return tool calls. Plan-mode and other flows that require a tool can receive ordinary text and proceed without the required action.

## Suggested fix

- Omit `params.tools` for `none`.
- Define whether a one-tool catalog plus output validation is sufficient for a specific choice; otherwise reject it.
- Reject unsupported `required` behavior before the request unless the selected Ollama version provides a checked contract.
- Record requested and effective tool choice in diagnostics.

## Acceptance criteria

- `none` sends no tools and rejects any impossible tool-call response.
- `required` and specific choices are either enforced and validated or fail locally with a clear provider-capability error.
- `auto` retains current catalog behavior.
- Plan-mode tests cover the required-tool path.
- Trace diagnostics show requested and effective choice.

## Duplicate check

Searched open and closed issues and pull requests for Ollama tool choice, `none`, required tools, specific tools, and plan-mode enforcement. No matching report was found.

turkonthelurk:
Closes https://github.com/tetsuo-ai/agenc-core/issues/2109

**Draft for review only.** Not to be merged by automation; a maintainer decides when and whether it lands. Opened from fork `turkonthelurk/agenc-core`. Nothing pushed to `main` on either repo. Based on upstream `main` only. Independent of #2110/#2590 and #2112/#2591.

## Problem

Ollama's native `/api/chat` accepts a tools list but does not expose OpenAI-style `tool_choice`. The adapter previously ignored `options.toolChoice`, so `none`, `required`, and a specific function all produced the same request body and could violate the caller's tool-use constraint.

## Change

| Choice | Contract |
| --- | --- |
| `auto` / omitted | Same selected catalog as before |
| `none` | Omit `params.tools` and the text-tool protocol catalog. Do not salvage text into a call. Reject any native tool-call response. |
| specific function | One-tool catalog plus output validation. Missing catalog entry, text-only, or wrong-tool responses fail. |
| `required` | Fail **before** the request with a provider-capability error (no checked native `tool_choice` contract). Plan-mode tests cover this path. |

Traces include `requestedToolChoice` and `effectiveToolChoice`. Metrics record the effective choice; `none` also sets `toolSuppressionReason: "tool_choice_none"`.

## Tests

Hermetic Ollama provider suite: **28/28** tool-choice tests; **253/253** full suite passed.

This does not merge.

Opened from fork `turkonthelurk/agenc-core`. Companion fork PR (to close): https://github.com/turkonthelurk/agenc-core/pull/31


7etsuo:
- [tool-choice.ts:128](https://github.com/tetsuo-ai/agenc-core/blob/e7468f6e8094557f0c47188e2a716dd5a3913c02/runtime/src/llm/providers/ollama/tool-choice.ts#L128) rejects `required`, but plan mode sets it whenever tools are present ([run-turn-sampling-request.ts:248](https://github.com/tetsuo-ai/agenc-core/blob/e7468f6e8094557f0c47188e2a716dd5a3913c02/runtime/src/session/run-turn-sampling-request.ts#L248)), so every plan-mode turn on Ollama fails before sending. Enforce `required` (full catalog, at least one advertised call) or normalize it to `auto`, and add a session-level plan-mode test.
- A truncated (`length`) reply under a named choice has its tool calls cleared, then fails the check at [adapter.ts:778](https://github.com/tetsuo-ai/agenc-core/blob/e7468f6e8094557f0c47188e2a716dd5a3913c02/runtime/src/llm/providers/ollama/adapter.ts#L778) and [adapter.ts:1255](https://github.com/tetsuo-ai/agenc-core/blob/e7468f6e8094557f0c47188e2a716dd5a3913c02/runtime/src/llm/providers/ollama/adapter.ts#L1255), so max-output recovery never runs. Skip the check when truncated and test both paths.
- Rebase after #2587 lands; both change adapter.ts.

PR #2633, opened 2026-09-20
Fix Meta completion validation and preserve context with output caps

signerless:
Meta streams could be accepted as successful without a terminal completion signal, leaving unfinished tool calls eligible for dispatch. Validate Meta SSE framing and require an explicit tool-call finish reason before accepting tool calls.

Configuring only an output-token limit also discarded the registered model context window and prevented session startup. Preserve built-in metadata beneath explicit overrides in both synchronous model selection and asynchronous startup, while keeping the configured output cap. Correct the Meta reasoning documentation.

Validation performed locally:
- 377 distinct focused provider, wire and metadata tests passed; the final metadata suite reran 92 of these tests.
- Runtime and test-support typechecks and the full build passed.
- Live Meta text streaming, tool execution and tool-result continuation passed. A Linux Desktop session saved the encrypted key, discovered five models and completed a chat with one provider dispatch capped at 512 output tokens.
- Rebased onto current main; the rebuilt runtime identifies this PR commit and the working tree is clean.

The audit used synthetic prompts. No credentials, account identifiers or private test profiles are included.


signerless:
Superseded. The same commit landed on main as tetsuo-ai/agenc-core#2632, `b16c502d`: `git diff origin/main pr2633` is empty, so this branch and main are identical.

The Desktop side is tetsuo-ai/agenc-desktop#321, and 0.1.2 pins this Core revision.

